PR #9, opened 2022-10-11
[FLINK-29443] Replicate packaging tests

darenwkt:
Added skeleton packaging tests to test if build jar contains expected files.

Files to look for currently are:
- META-INF/

### DIFF
--- a/flink-sql-connector-aws-kinesis-firehose/src/test/java/org/apache/flink/connector/firehose/PackagingITCase.java
+++ b/flink-sql-connector-aws-kinesis-firehose/src/test/java/org/apache/flink/connector/firehose/PackagingITCase.java
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.connector.firehose;
+
+import org.apache.flink.packaging.PackagingTestUtils;
+import org.apache.flink.table.factories.Factory;
+import org.apache.flink.test.resources.ResourceTestUtils;
+
+import org.junit.jupiter.api.Test;
+
+import java.nio.file.Path;
+import java.util.Arrays;
+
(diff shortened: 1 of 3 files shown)


dannycranmer:
@darenwkt the checks are failing on CI

dannycranmer:
@darenwkt please squash commits and fix commit message to comply with the [contribution guide](https://flink.apache.org/contributing/contribute-documentation.html#submit-your-contribution)

flink-connector-dynamodb/pom.xml
```diff
@@ -33,6 +33,15 @@ under the License.
     <artifactId>flink-connector-dynamodb</artifactId>
     <name>Flink : Connectors : Amazon DynamoDB</name>
 
+	<dependencies>
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-test-utils</artifactId>
+			<version>1.16-SNAPSHOT</version>
```
dannycranmer, line 40: Why are we depending on a snapshot version and not the parent Flink version? Can this dependency not be managed from the dynamodb parent pom?

flink-connector-dynamodb/src/test/java/org/apache/flink/connectors/dynamodb/PackagingITCase.java
```diff
@@ -0,0 +1,19 @@
...
+import java.nio.file.Path;
+import java.util.Arrays;
+
+/** Test to verify contents of packaged jar. */
+public class PackagingITCase {
+    @Test
+    void testPackaging() throws Exception {
+        final Path jar = ResourceTestUtils.getResource(".*/flink-connector-dynamodb-[^/]*\\.jar");
+
+        PackagingTestUtils.assertJarContainsOnlyFilesMatching(jar, Arrays.asList("META-INF/"));
```
dannycranmer, line 17: This will break the build for the main code contribution I assume. Can we relax this check to support the intended result? We expect class file etc...

flink-sql-connector-dynamodb/pom.xml
```diff
@@ -51,6 +51,13 @@ under the License.
 			<artifactId>flink-connector-dynamodb</artifactId>
 			<version>${project.version}</version>
 		</dependency>
+
+		<dependency>
+			<groupId>org.apache.flink</groupId>
+			<artifactId>flink-test-utils</artifactId>
+			<version>1.16-SNAPSHOT</version>
```
dannycranmer, line 58: Same as above

dannycranmer (changes requested):

flink-connector-dynamodb/pom.xml
darenwkt: Had offline discussion with Danny, we will use 1.16-SNAPSHOT version as there is no released version for flink-test-utils and flink-ci-tools yet. 

Todo: Once 1.16 is release, we will upgrade the version of dependencies in here to released verison

flink-connector-dynamodb/src/test/java/org/apache/flink/connectors/dynamodb/PackagingITCase.java
darenwkt: I have relax the regex to match any dir name prefix with "flink-"

dannycranmer (commented):
LGTM

PR #22, opened 2022-11-14
[FLINK-29688] Add test to detect changes in DynamoDB model

dannycranmer:

Add test to detect changes in DynamoDB model

## Brief change log

* Add test to detect changes in DynamoDB model

## Verifying this change

Tests pass

  - Dependencies (does it add or upgrade a dependency): no
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: no
  - The serializers: no
  - The runtime per-record code paths (performance sensitive): no
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Kubernetes/Yarn, ZooKeeper: no
  - The S3 file system connector: no

## Documentation

  - Does this pull request introduce a new feature? no
  - If yes, how is the feature documented? n/a


hlteoh37 (approved):
LGTM. Just wonder if we should be a little tighter on the contract (e.g. field types). However, approving, since it is more likely that when contract changes, fields are added, rather than modified (for backwards compatibility)

PR #36, opened 2022-12-12
[FLINK-30378][docs] Use modified sql_connector_download_table shortcode

zentol:
The updated shortcode proposed in FLINK-30378 allows the connector to provide the sql connector data.


docs/data/firehose.yml
```diff
@@ -0,0 +1,3 @@
+variants:
+  - maven: flink-connector-kinesis-firehose
```
dannycranmer, line 2: This should be `flink-connector-aws-kinesis-firehose`

dannycranmer (changes requested):

PR #38, opened 2022-12-13
[hotfix][ci] Sync CI setup & Reuse workflow

zentol:
Migrates this repo to the common CI setup and newly reusable workflow. AFAICT the main change is not having 2 separate steps for normal/e2e tests.
Please let me know if that's a significant problem for you.


zentol:
The test failure is likely caused by an actual packaging problem, that was previously masked because you were using Maven 3.8.5 which has known issues w.r.t. dependency propagation.
Actually, you probably shouldn't be using Maven 3.3+ due to the usual shading issues yet unless you cherry-pick the relevant changes from https://github.com/apache/flink/pull/21349.

PR #39, opened 2022-12-14
[FLINK-30418] Implement synchronous KinesisClient in EFO

darenwkt:
Implemented synchronous KinesisClient as described in https://issues.apache.org/jira/browse/FLINK-30418.

KinesisProxyV2 supports the following operations:

    subscribeToShard
    describeStreamSummary
    describeStreamConsumer
    registerStreamConsumer
    deregisterStreamConsumer

 

All the above operations are currently performed via an AsyncHTTPClient. However, only subsribeToShard needs to be asynchronous, the rest are called synchronously. This PR updates all the operation to use a SyncHTTPClient except subscribeToShard.

Change Summary:

    subscribeToShard: async (no change)
    describeStreamSummary: async -> sync
    describeStreamConsumer: async -> sync
    registerStreamConsumer: async -> sync
    deregisterStreamConsumer: async -> sync

Tested in local environment with a sample input and output Kinesis Stream on local flink cluster.


dannycranmer:
@darenwkt did you verify this works using EFO? There are no e2e tests for EFO since Kinesalite does not support it

flink-connector-aws-base/pom.xml
```diff
@@ -52,6 +52,11 @@ under the License.
             <artifactId>netty-nio-client</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>apache-client</artifactId>
+        </dependency>
```
dannycranmer, line 58: We will need to ensure this is shaded correctly and NOTICE files are updated

flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/util/AWSAsyncSinkUtil.java
```diff
@@ -161,4 +163,86 @@ S createAwsAsyncClient(
...
+     * @param configProps configuration properties
+     * @param clientConfiguration the AWS SDK v2 config to instantiate the client
+     * @param httpClient the underlying HTTP client used to talk to AWS
+     * @return a new AWS Sync Client
+     */
+    public static <
+                    S extends SdkClient,
+                    T extends
+                            AwsSyncClientBuilder<? extends T, S> & AwsClientBuilder<? extends T, S>>
+            S createAwsSyncClient(
```
dannycranmer, line 202: Why do we need so many overloaded versions of this? Looks like only 1 is used

flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/util/AWSAsyncSinkUtil.java
```diff
@@ -161,4 +163,86 @@ S createAwsAsyncClient(
...
+        final ClientOverrideConfiguration overrideConfiguration =
+                createClientOverrideConfiguration(
+                        clientConfiguration,
+                        ClientOverrideConfiguration.builder(),
+                        flinkUserAgentPrefix);
+
+        return createAwsSyncClient(configProps, clientBuilder, httpClient, overrideConfiguration);
+    }
+
+    @VisibleForTesting
```
dannycranmer, line 224: Where is this tested such that it requires this scope?

flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/util/AWSGeneralUtil.java
```diff
@@ -301,6 +303,58 @@ public static SdkAsyncHttpClient createAsyncHttpClient(
...
+                                clientConfiguration.put(
+                                        SdkHttpConfigurationOption.TRUST_ALL_CERTIFICATES, bool));
+
+        Optional.ofNullable(configProperties.getProperty(AWSConfigConstants.HTTP_PROTOCOL_VERSION))
+                .map(Protocol::valueOf)
+                .ifPresent(
+                        protocol ->
+                                clientConfiguration.put(
+                                        SdkHttpConfigurationOption.PROTOCOL, protocol));
+
```
dannycranmer, line 348: Seems like this is duplicated for Async/Sync. Can we extract to a common method?

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/proxy/KinesisProxyV2Factory.java
```diff
@@ -55,25 +58,39 @@ public static KinesisProxyV2Interface createKinesisProxyV2(final Properties conf
...
+        final KinesisClient client =
+                AWSAsyncSinkUtil.createAwsSyncClient(
+                        clientProperties,
+                        httpClient,
+                        KinesisClient.builder(),
+                        KinesisStreamsConfigConstants.BASE_KINESIS_USER_AGENT_PREFIX_FORMAT,
+                        KinesisStreamsConfigConstants.KINESIS_CLIENT_USER_AGENT_PREFIX);
+
+        return new KinesisProxyV2(
+                asyncClient, client, asyncHttpClient, httpClient, configuration, BACKOFF);
```
dannycranmer, line 93: As we discussed, we should split `KinesisProxyV2` so we do not need to create both client. In the case of registerStreamConsumer it will not use the async client and it is expensive to create a thread pool, then trash it. We should only create the async client when we actually need to use it

dannycranmer (changes requested):

darenwkt:
> @darenwkt did you verify this works using EFO? There are no e2e tests for EFO since Kinesalite does not support it

Yes, I verified by running EFOConsumer examples from https://github.com/aws-samples/amazon-kinesis-data-analytics-examples/blob/master/EfoConsumer/pom.xml against my local Flink 1.16.0 cluster.

flink-connector-aws-base/pom.xml
darenwkt: Verified that apache-client is in the installed JAR.

Screenshot 2022-12-16 at 19 51 45


Verified that NOTICE files are updated.

```

- software.amazon.awssdk:apache-client:2.17.247
- software.amazon.awssdk:netty-nio-client:2.17.247
```

flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/util/AWSAsyncSinkUtil.java
darenwkt: Done, have cleaned the rest

flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/util/AWSAsyncSinkUtil.java
darenwkt: Done, have clean this up

flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/util/AWSGeneralUtil.java
darenwkt: Done, extracted to common method

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/proxy/KinesisProxyV2Factory.java
darenwkt: Thank you for clarifying, I have split `KinesisProxyV2` to `KinesisProxySyncV2` and `KinesisProxyAsyncV2`. 

For StreamRegistrar related operations (i.e registerStreamConsumer), only `KinesisProxySyncV2` will be created. 
For EFO related operations (i.e subsribeToShard), only `KinesisProxyAsyncV2` will be created.

flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/util/AWSAsyncSinkUtil.java
```diff
@@ -161,4 +163,47 @@ S createAwsAsyncClient(
                 .region(getRegion(configProps))
                 .build();
     }
+
+    /**
+     * @param configProps configuration properties
```
dannycranmer, line 168: This javadoc is missing a comment.

flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/util/AWSAsyncSinkUtil.java
```diff
@@ -161,4 +163,47 @@ S createAwsAsyncClient(
...
+                        clientConfiguration,
+                        ClientOverrideConfiguration.builder(),
+                        flinkUserAgentPrefix);
+
+        if (configProps.containsKey(AWSConfigConstants.AWS_ENDPOINT)) {
+            final URI endpointOverride =
+                    URI.create(configProps.getProperty(AWSConfigConstants.AWS_ENDPOINT));
+            clientBuilder.endpointOverride(endpointOverride);
+        }
+
```
dannycranmer, line 201: This code is duplicated in this file, can you extract to a method please

flink-connector-kinesis/pom.xml
```diff
@@ -256,6 +256,11 @@ under the License.
             <artifactId>netty-nio-client</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>apache-client</artifactId>
+        </dependency>
+
```
dannycranmer, line 263: Why is this needed explicitly in here and `aws-base`?

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/internals/publisher/fanout/StreamConsumerRegistrar.java
```diff
@@ -77,8 +79,8 @@ public StreamConsumerRegistrar(
      * @throws InterruptedException
      */
     public String registerStreamConsumer(final String stream, final String streamConsumerName)
-            throws ExecutionException, InterruptedException {
-        LOG.debug("Registering stream consumer - {}::{}", stream, streamConsumerName);
+            throws Exception {
```
dannycranmer, line 82: Why did you make this change? This will possible change the error handling behaviour

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/internals/publisher/fanout/StreamConsumerRegistrar.java
```diff
@@ -77,8 +79,8 @@ public StreamConsumerRegistrar(
...
+        LOG.warn("Registering stream consumer - {}::{}", stream, streamConsumerName);
```
dannycranmer, line 83: Why is this a warning?

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/internals/publisher/fanout/StreamConsumerRegistrar.java
```diff
@@ -90,7 +92,7 @@ public String registerStreamConsumer(final String stream, final String streamCon
                 kinesisProxyV2Interface.describeStreamSummary(stream);
         String streamArn = describeStreamSummaryResponse.streamDescriptionSummary().streamARN();
 
-        LOG.debug("Found stream ARN - {}", streamArn);
+        LOG.warn("Found stream ARN - {}", streamArn);
```
dannycranmer, line 95: Why is this a warning?

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/internals/publisher/fanout/StreamConsumerRegistrar.java
```diff
@@ -243,28 +244,28 @@ private void waitForConsumerToDeregister(
...
-            throws InterruptedException, ExecutionException {
+            final Callable<DescribeStreamConsumerResponse> responseSupplier) throws Exception {
         DescribeStreamConsumerResponse response;
 
         try {
-            response = responseSupplier.get();
-        } catch (ExecutionException ex) {
+            response = responseSupplier.call();
+        } catch (Exception ex) {
+            LOG.warn("describeStreamConsumer caught ExecutionException: {}", ex);
```
dannycranmer, line 267: `caught ExecutionException` not necessarily now you are catching `Exception`. This change is slightly worrying, what happens now if an `InterruptedException` is thrown?

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/proxy/KinesisProxySyncV2.java
```diff
@@ -148,11 +139,10 @@ public RegisterStreamConsumerResponse registerStreamConsumer(
                         .consumerName(consumerName)
                         .build();
 
+        LOG.warn("registerStreamConsumer with arn: {}, consumerName: {}", streamArn, consumerName);
```
dannycranmer, line 142: Why is this a warning?

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/proxy/KinesisProxySyncV2.java
```diff
@@ -114,6 +104,7 @@ public DescribeStreamConsumerResponse describeStreamConsumer(
                         .streamARN(streamArn)
                         .consumerName(consumerName)
                         .build();
+        LOG.warn("describeStreamConsumer with arn: {}, consumerName: {}", streamArn, consumerName);
```
dannycranmer, line 107: Why is this a warning? (did you leave dev debug logs in?)

flink-connector-kinesis/src/test/java/org/apache/flink/streaming/connectors/kinesis/proxy/KinesisProxyAsyncV2Test.java
```diff
@@ -0,0 +1,171 @@
...
+import static org.apache.flink.streaming.connectors.kinesis.config.ConsumerConfigConstants.STREAM_DESCRIBE_BACKOFF_BASE;
+import static org.apache.flink.streaming.connectors.kinesis.config.ConsumerConfigConstants.STREAM_DESCRIBE_BACKOFF_EXPONENTIAL_CONSTANT;
+import static org.apache.flink.streaming.connectors.kinesis.config.ConsumerConfigConstants.STREAM_DESCRIBE_BACKOFF_MAX;
+import static org.apache.flink.streaming.connectors.kinesis.config.ConsumerConfigConstants.STREAM_DESCRIBE_RETRIES;
+import static org.apache.flink.streaming.connectors.kinesis.config.ConsumerConfigConstants.SUBSCRIBE_TO_SHARD_BACKOFF_BASE;
+import static org.apache.flink.streaming.connectors.kinesis.config.ConsumerConfigConstants.SUBSCRIBE_TO_SHARD_BACKOFF_EXPONENTIAL_CONSTANT;
+import static org.apache.flink.streaming.connectors.kinesis.config.ConsumerConfigConstants.SUBSCRIBE_TO_SHARD_BACKOFF_MAX;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
```
dannycranmer, line 56: Mockito is [banned from use](https://flink.apache.org/contributing/code-style-and-quality-common.html#avoid-mockito---use-reusable-test-implementations) in Flink. I know we do already use it in places but we should not use for new tests. Can we refactor to reusable test implementations?

dannycranmer (changes requested):

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/internals/publisher/fanout/StreamConsumerRegistrar.java
darenwkt: Done, fixed this

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/internals/publisher/fanout/StreamConsumerRegistrar.java
darenwkt: Done, fixed this

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/proxy/KinesisProxySyncV2.java
darenwkt: Yes, my apology for missing this in post-cleanup, doing a round of check again to remove them

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/internals/publisher/fanout/StreamConsumerRegistrar.java
darenwkt: The main reason of the change is that we are using SyncClient for registerStreamConsumer, so it is not a FutureTask and it won't throw ExecutionException. However, I agree that InterruptedException should be readded as it might be thrown during registrationBackoff

flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/util/AWSAsyncSinkUtil.java
darenwkt: Done added

flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/util/AWSAsyncSinkUtil.java
darenwkt: Done refactored

flink-connector-kinesis/pom.xml
darenwkt: Got it, removing this and keep it in aws-base

flink-connector-kinesis/src/test/java/org/apache/flink/streaming/connectors/kinesis/proxy/KinesisProxyAsyncV2Test.java
darenwkt: Have refactored this away from mockito

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/internals/publisher/fanout/StreamConsumerRegistrar.java
vahmed-hamdy: I agree with @dannycranmer , we would better not generalize the Exception.

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/internals/publisher/fanout/StreamConsumerRegistrar.java
```diff
@@ -218,7 +219,7 @@ private void waitForConsumerToDeregister(
             @Nullable final DescribeStreamConsumerResponse describeStreamConsumerResponse,
             final String streamConsumerArn,
             final int initialAttempt)
-            throws InterruptedException, ExecutionException {
+            throws Exception {
```
vahmed-hamdy, line 222: Same goes here

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/proxy/KinesisProxyAsyncV2Interface.java
```diff
@@ -0,0 +1,41 @@
...
+ * same AWS service region.
+ */
+@Internal
+public interface KinesisProxyAsyncV2Interface {
+
+    CompletableFuture<Void> subscribeToShard(
+            SubscribeToShardRequest request, SubscribeToShardResponseHandler responseHandler);
+
+    /** Destroy any open resources used by the factory. */
+    default void close() {
```
vahmed-hamdy, line 38: Why do we provide a default no-op method?

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/internals/publisher/fanout/StreamConsumerRegistrar.java
darenwkt: I think we can generalise it here because previously ExecutionException, InterruptedException were thrown from a CompletableFuture get method. However, we are no longer using Future here as registerStreamConsumer is called synchronously now (via Callable call method) instead of async. Unlike CompletableFuture, Callable call method throws Exception instead of Execution + InterruptedException.

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/internals/publisher/fanout/StreamConsumerRegistrar.java
darenwkt: Same reply as above: https://github.com/apache/flink-connector-aws/pull/39/files/de6c5579a3b1d4b59938a7b6c3fcdc56d4ef88d4#r1058982327

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/proxy/KinesisProxyAsyncV2Interface.java
darenwkt: In the implementation (i.e KinesisProxyAsyncV2.java), close() is overriden and used to close any open HttpClient gracefully. 

I think it's ok to put as default for cases that do not require closing any resources. But do let me know if you think I should remove the default so that any implementation have to explicitly implement it.

dannycranmer:
@darenwkt please repush with correct component tag in commit message, ie `[FLINK-XXXX][Connectors/Kinesis] ...`

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/internals/publisher/fanout/StreamConsumerRegistrar.java
dannycranmer: My concerns is the Exception is a wider catch bucket than before. However given `ExecutionException, InterruptedException` would have caught the majority and the parent wraps in a `FlinkKinesisStreamConsumerRegistrarException`, it looks ok

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/proxy/KinesisProxyAsyncV2.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Kinesis proxy implementation using AWS SDK v2.x - a utility class that is used as a proxy to make
+ * calls to AWS Kinesis for several EFO (Enhanced Fan Out) functions, such as de-/registering stream
+ * consumers, subscribing to a shard and receiving records from a shard.
+ */
+@Internal
+public class KinesisProxyAsyncV2 implements KinesisProxyAsyncV2Interface {
+
+    private static final Logger LOG = LoggerFactory.getLogger(KinesisProxyAsyncV2.class);
```
dannycranmer, line 41: This is unused

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/proxy/KinesisProxyAsyncV2Interface.java
dannycranmer: Seems like a shortcut for testing. I would rather move it to `KinesisProxyAsyncV2InterfaceAdapter` to force developers to think about `close()` in src. We missed this before and it caused issues!

dannycranmer (changes requested):

flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/util/AWSAsyncSinkUtil.java
```diff
@@ -161,4 +160,64 @@ S createAwsAsyncClient(
...
+     * @param clientBuilder httpClientBuilder to build the underlying HTTP client
+     * @param awsUserAgentPrefixFormat user agent prefix for Flink
+     * @param awsClientUserAgentPrefix user agent prefix for kinesis client
+     * @return a new AWS Sync Client
+     */
+    public static <
+                    S extends SdkClient,
+                    T extends
+                            AwsSyncClientBuilder<? extends T, S> & AwsClientBuilder<? extends T, S>>
+            S createAwsSyncClient(
```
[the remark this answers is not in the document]
darenwkt: I think refactoring it to AWSGeneralUtil will be messy as some Client Creation logic will have to be refactored to AWSGeneralUtil. Can I suggest renaming `AWSAsyncSinkUtil` to `AWSClientUtil` instead? This will be more tidy as all the utils to create client are in this class.

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/proxy/KinesisProxyAsyncV2.java
darenwkt: Have removed, thank you

flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/proxy/KinesisProxyAsyncV2Interface.java
darenwkt: Got it, I have removed default and moved this to KinesisProxyAsyncV2InterfaceAdapter and KinesisProxySyncV2InterfaceAdapter to implement

darenwkt:
@dannycranmer, @vahmed-hamdy I have addressed the comments and retested locally against a local Flink 1.16 cluster and Kinesis EFO Stream. Appreciate if you could review again, thank you

flink-connector-aws-base/pom.xml
dannycranmer: NOTICE files do not update automatically, so either:
- It is already included and this is not necessary, or
- The NOTICE files need updating

Since flink-connector-aws-base is common to all connectors we will need to update all the NOTICE files I believe.

dannycranmer: Yes, I already see it. For all connectors, thanks

dannycranmer (approved):

PR #42, opened 2022-12-22
[FLINK-30229] Add SQL IT case for DynamoDB Sink

darenwkt:
Added integration test against DynamoDBLocal running in Docker with dummy data generated from datagen SQL connector.


flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkITCase.java
```diff
@@ -0,0 +1,195 @@
...
+                + "WITH (\n"
+                + "    'connector' = 'datagen',\n"
+                + "    'number-of-rows' = '"
+                + expectedNumOfElements
+                + "'\n"
+                + ")\n"
+                + "LIKE dynamo_db_table (EXCLUDING ALL);";
+    }
+
+    private static String getCreateTableStmt() {
```
hlteoh37, line 132: Q: can we move this to a java resource? same for the other statements

flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkITCase.java
```diff
@@ -0,0 +1,195 @@
...
+
+        dynamoDBHelpers.createTable(testTableName, PARTITION_KEY, SORT_KEY);
+        StreamTableEnvironment streamTableEnvironment =
+                StreamTableEnvironment.create(env, EnvironmentSettings.newInstance().build());
+
+        final String createTableStmt = getCreateTableStmt();
+        streamTableEnvironment.executeSql(createTableStmt);
+
+        final String datagenStmt = getDatagenStmt(expectedNumOfElements);
+        streamTableEnvironment.executeSql(datagenStmt);
```
hlteoh37, line 111: can we also use Table API for this?

flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkITCase.java
vahmed-hamdy: +1 for resource usage

flink-connector-dynamodb/pom.xml
```diff
@@ -132,5 +132,12 @@ under the License.
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-table-planner-loader</artifactId>
```
vahmed-hamdy, line 137: previously we had separate modules for these IT tests, they were more of E2E tests for uber sql connector. since we are not doing this with this approach are we intending to have another E2E test?

flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkITCase.java
```diff
@@ -0,0 +1,195 @@
...
+
+        Table resultTable = streamTableEnvironment.sqlQuery("SELECT * FROM datagen");
+
+        resultTable.executeInsert("dynamo_db_table").await();
+
+        Assertions.assertThat(dynamoDBHelpers.getItemsCount(testTableName))
+                .isEqualTo(expectedNumOfElements);
```
vahmed-hamdy, line 118: Refering to Firehose test, we used to see intermittent failures due to delays between sending records downstream and observing them from LocalStack, is this not needed for DDB?

flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkITCase.java
darenwkt: Agreed, will update to use TableAPI and TableEnvironment for this

flink-connector-dynamodb/pom.xml
darenwkt: Yes, I was thinking that E2E test will be testing against an actual DynamoDB whereas this IT case is testing against a local DynamoDB container.

flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkITCase.java
darenwkt: The line before is a synchronous operation "resultTable.executeInsert("dynamo_db_table").await();" which should wait until all records are successfully sent. So the delay shouldn't cause us failures here unless the await timeouts

flink-connector-dynamodb/pom.xml
vahmed-hamdy: I am happy to follow this approach if we intend to have E2E tests, they need not to be against an actual DDB (this is preferable ofc) but we need to verify that the connector works when packaging the uber jar to sql client.

KDS tests for reference
https://github.com/apache/flink-connector-aws/blob/2a24e7c1e8d2a8f613838fd0419177d4f96991bf/flink-connector-aws-e2e-tests/flink-connector-aws-kinesis-streams-e2e-tests/pom.xml#L99

vahmed-hamdy:
LGTM, If we can create a JIRA for an E2E tests that would be great as well!

vahmed-hamdy (approved):

dannycranmer (approved):

PR #49, opened 2023-01-25
[FLINK-24438] Add Kinesis connector using FLIP-27

hlteoh37:
- Implement FlinkKinesisSource using the new FLIP-27 Source interface

* Dependencies (does it add or upgrade a dependency): no
* The public API, i.e., is any changed class annotated with `@Public(Evolving)`: yes
* The serializers: no
* The runtime per-record code paths (performance sensitive): no
* Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Kubernetes/Yarn, ZooKeeper: no
* The S3 file system connector: no

## Documentation
* Does this pull request introduce a new feature? no
* If yes, how is the feature documented? n/a

### DIFF
--- a/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
+++ b/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
@@ -1,0 +1,205 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.connector.kinesis.source;
+
+import org.apache.flink.annotation.Experimental;
+import org.apache.flink.api.common.serialization.DeserializationSchema;
+import org.apache.flink.api.connector.source.Boundedness;
+import org.apache.flink.api.connector.source.Source;
+import org.apache.flink.api.connector.source.SourceReader;
+import org.apache.flink.api.connector.source.SourceReaderContext;
+import org.apache.flink.api.connector.source.SplitEnumerator;
+import org.apache.flink.api.connector.source.SplitEnumeratorContext;
+import org.apache.flink.configuration.ConfigurationUtils;
+import org.apache.flink.connector.aws.util.AWSClientUtil;
+import org.apache.flink.connector.aws.util.AWSGeneralUtil;
+import org.apache.flink.connector.base.source.reader.RecordsWithSplitIds;
+import org.apache.flink.connector.base.source.reader.fetcher.SingleThreadFetcherManager;
+import org.apache.flink.connector.base.source.reader.synchronization.FutureCompletingBlockingQueue;
+import org.apache.flink.connector.kinesis.sink.KinesisStreamsConfigConstants;
+import org.apache.flink.connector.kinesis.source.enumerator.KinesisShardAssigner;
+import org.apache.flink.connector.kinesis.source.enumerator.KinesisStreamsSourceEnumerator;
+import org.apache.flink.connector.kinesis.source.enumerator.KinesisStreamsSourceEnumeratorState;
+import org.apache.flink.connector.kinesis.source.enumerator.KinesisStreamsSourceEnumeratorStateSerializer;
+import org.apache.flink.connector.kinesis.source.proxy.KinesisStreamProxy;
+import org.apache.flink.connector.kinesis.source.reader.KinesisStreamsRecordEmitter;
+import org.apache.flink.connector.kinesis.source.reader.KinesisStreamsSourceReader;
+import org.apache.flink.connector.kinesis.source.reader.PollingKinesisShardSplitReader;
+import org.apache.flink.connector.kinesis.source.serialization.KinesisDeserializationSchema;
+import org.apache.flink.connector.kinesis.source.split.KinesisShardSplit;
+import org.apache.flink.connector.kinesis.source.split.KinesisShardSplitSerializer;
+import org.apache.flink.core.io.SimpleVersionedSerializer;
+import org.apache.flink.metrics.MetricGroup;
+import org.apache.flink.util.Preconditions;
+import org.apache.flink.util.UserCodeClassLoader;
+
+import software.amazon.awssdk.http.SdkHttpClient;
+import software.amazon.awssdk.http.apache.ApacheHttpClient;
+import software.amazon.awssdk.services.kinesis.KinesisClient;
+import software.amazon.awssdk.services.kinesis.model.Record;
+import software.amazon.awssdk.utils.AttributeMap;
+
+import java.util.Properties;
+import java.util.function.Supplier;
+
+/**
+ * The {@link KinesisStreamsSource} is an exactly-once parallel streaming data source that
+ * subscribes to a single AWS Kinesis data stream. It is able to handle resharding of streams, and
+ * stores its current progress in Flink checkpoints. The source will read in data from the Kinesis
+ * Data stream, deserialize it using the provided {@link DeserializationSchema}, and emit the record
+ * into the Flink job graph.
+ *
+ * <p>Exactly-once semantics. To leverage Flink's checkpointing mechanics for exactly-once stream
+ * processing, the Kinesis Source is implemented with the AWS Java SDK, instead of the officially
+ * recommended AWS Kinesis Client Library. The source will store its current progress in Flink
+ * checkpoint/savepoint, and will pick up from where it left off upon restore from the
+ * checkpoint/savepoint.
+ *
+ * <p>Initial starting points. The Kinesis Streams Source supports reads starting from TRIM_HORIZON,
+ * LATEST, and AT_TIMESTAMP.
+ *
+ * @param <T> the data type emitted by the source
+ */
+@Experimental
+public class KinesisStreamsSource<T>
+        implements Source<T, KinesisShardSplit, KinesisStreamsSourceEnumeratorState> {
+
+    private final String streamArn;
+    private final Properties consumerConfig;
+    private final KinesisDeserializationSchema<T> deserializationSchema;
+    private final KinesisShardAssigner kinesisShardAssigner;
+
+    KinesisStreamsSource(
+            String streamArn,
+            Properties consumerConfig,
+            KinesisDeserializationSchema<T> deserializationSchema,
+            KinesisShardAssigner kinesisShardAssigner) {
+        Preconditions.checkNotNull(streamArn);
+        Preconditions.checkArgument(!streamArn.isEmpty(), "stream ARN cannot be empty string");
+        Preconditions.checkNotNull(consumerConfig);
+        Preconditions.checkNotNull(deserializationSchema);
+        Preconditions.checkNotNull(kinesisShardAssigner);
+        this.streamArn = streamArn;
+        this.consumerConfig = consumerConfig;
+        this.deserializationSchema = deserializationSchema;
+        this.kinesisShardAssigner = kinesisShardAssigner;
+    }
+
+    /**
+     * Create a {@link KinesisStreamsSourceBuilder} to allow the fluent construction of a new {@link
+     * KinesisStreamsSource}.
+     *
+     * @param <T> type of records being read
+     * @return {@link KinesisStreamsSourceBuilder}
+     */
+    public static <T> KinesisStreamsSourceBuilder<T> builder() {
+        return new KinesisStreamsSourceBuilder<>();
+    }
+
+    @Override
+    public Boundedness getBoundedness() {
+        return Boundedness.CONTINUOUS_UNBOUNDED;
+    }
+
+    @Override
+    public SourceReader<T, KinesisShardSplit> createReader(SourceReaderContext readerContext)
+            throws Exception {
+        setUpDeserializationSchema(readerContext);
+
+        FutureCompletingBlockingQueue<RecordsWithSplitIds<Record>> elementsQueue =
+                new FutureCompletingBlockingQueue<>();
+        // We create a new stream proxy for each split reader since they have their own independent
+        // lifecycle.
+        Supplier<PollingKinesisShardSplitReader> splitReaderSupplier =
+                () -> new PollingKinesisShardSplitReader(createKinesisStreamProxy(consumerConfig));
+        KinesisStreamsRecordEmitter<T> recordEmitter =
+                new KinesisStreamsRecordEmitter<>(deserializationSchema);
+
+        return new KinesisStreamsSourceReader<>(
+                elementsQueue,
+                new SingleThreadFetcherManager<>(elementsQueue, splitReaderSupplier::get),
+                recordEmitter,
+                ConfigurationUtils.createConfiguration(consumerConfig),
+                readerContext);
+    }
+
+    @Override
+    public SplitEnumerator<KinesisShardSplit, KinesisStreamsSourceEnumeratorState> createEnumerator(
+            SplitEnumeratorContext<KinesisShardSplit> enumContext) throws Exception {
+        return restoreEnumerator(enumContext, null);
+    }
+
+    @Override
+    public SplitEnumerator<KinesisShardSplit, KinesisStreamsSourceEnumeratorState>
+            restoreEnumerator(
+                    SplitEnumeratorContext<KinesisShardSplit> enumContext,
+                    KinesisStreamsSourceEnumeratorState checkpoint)
+                    throws Exception {
+        return new KinesisStreamsSourceEnumerator(
+                enumContext,
+                streamArn,
+                consumerConfig,
+                createKinesisStreamProxy(consumerConfig),
+                kinesisShardAssigner,
+                checkpoint);
+    }
+
+    @Override
+    public SimpleVersionedSerializer<KinesisShardSplit> getSplitSerializer() {
+        return new KinesisShardSplitSerializer();
+    }
+
+    @Override
+    public SimpleVersionedSerializer<KinesisStreamsSourceEnumeratorState>
+            getEnumeratorCheckpointSerializer() {
+        return new KinesisStreamsSourceEnumeratorStateSerializer(new KinesisShardSplitSerializer());
+    }
+
+    private KinesisStreamProxy createKinesisStreamProxy(Properties consumerConfig) {
+        SdkHttpClient httpClient =
+                AWSGeneralUtil.createSyncHttpClient(
+                        AttributeMap.builder().build(), ApacheHttpClient.builder());
+
+        AWSGeneralUtil.validateAwsCredentials(consumerConfig);
+        KinesisClient kinesisClient =
+                AWSClientUtil.createAwsSyncClient(
+                        consumerConfig,
+                        httpClient,
+                        KinesisClient.builder(),
+                        KinesisStreamsConfigConstants.BASE_KINESIS_USER_AGENT_PREFIX_FORMAT,
+                        KinesisStreamsConfigConstants.KINESIS_CLIENT_USER_AGENT_PREFIX);
+        return new KinesisStreamProxy(kinesisClient, httpClient);
+    }
+
+    private void setUpDeserializationSchema(SourceReaderContext sourceReaderContext)
+            throws Exception {
+        deserializationSchema.open(
+                new DeserializationSchema.InitializationContext() {
+                    @Override
+                    public MetricGroup getMetricGroup() {
+                        return sourceReaderContext.metricGroup().addGroup("deserializer");
+                    }
+
+                    @Override
+                    public UserCodeClassLoader getUserCodeClassLoader() {
+                        return sourceReaderContext.getUserCodeClassLoader();
+                    }
+                });
+    }
+}
--- a/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSourceBuilder.java
+++ b/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSourceBuilder.java
@@ -1,0 +1,92 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.connector.kinesis.source;
+
+import org.apache.flink.annotation.Experimental;
+import org.apache.flink.api.common.serialization.DeserializationSchema;
+import org.apache.flink.connector.kinesis.source.enumerator.KinesisShardAssigner;
+import org.apache.flink.connector.kinesis.source.enumerator.assigner.ShardAssignerFactory;
+import org.apache.flink.connector.kinesis.source.enumerator.assigner.UniformShardAssigner;
+import org.apache.flink.connector.kinesis.source.serialization.KinesisDeserializationSchema;
+
+import java.util.Properties;
+
+/**
+ * Builder to construct the {@link KinesisStreamsSource}.
+ *
+ * <p>The following example shows the minimum setup to create a {@link KinesisStreamsSource} that
+ * reads String values from a Kinesis Data Streams stream with ARN of
+ * arn:aws:kinesis:us-east-1:012345678901:stream/your_stream_name.
+ *
+ * <pre>{@code
+ * KinesisStreamsSource<String> kdsSource =
+ *                 KinesisStreamsSource.<String>builder()
+ *                         .setStreamArn("arn:aws:kinesis:us-east-1:012345678901:stream/your_stream_name")
+ *                         .setDeserializationSchema(new SimpleStringSchema())
+ *                         .build();
+ * }</pre>
+ *
+ * <p>If the following parameters are not set in this builder, the following defaults will be used:
+ *
+ * <ul>
+ *   <li>{@code kinesisShardAssigner} will be {@link UniformShardAssigner}
+ * </ul>
+ *
+ * @param <T> type of elements that should be read from the source stream
+ */
+@Experimental
+public class KinesisStreamsSourceBuilder<T> {
+    private String streamArn;
+    private Properties consumerConfig;
+    private KinesisDeserializationSchema<T> deserializationSchema;
+    private KinesisShardAssigner kinesisShardAssigner = ShardAssignerFactory.uniformShardAssigner();
+
+    public KinesisStreamsSourceBuilder<T> setStreamArn(String streamArn) {
+        this.streamArn = streamArn;
+        return this;
+    }
+
+    public KinesisStreamsSourceBuilder<T> setConsumerConfig(Properties consumerConfig) {
+        this.consumerConfig = consumerConfig;
+        return this;
+    }
+
+    public KinesisStreamsSourceBuilder<T> setDeserializationSchema(
+            KinesisDeserializationSchema<T> deserializationSchema) {
+        this.deserializationSchema = deserializationSchema;
+        return this;
+    }
+
+    public KinesisStreamsSourceBuilder<T> setDeserializationSchema(
+            DeserializationSchema<T> deserializationSchema) {
+        this.deserializationSchema = KinesisDeserializationSchema.of(deserializationSchema);
+        return this;
+    }
+
+    public KinesisStreamsSourceBuilder<T> setKinesisShardAssigner(
+            KinesisShardAssigner kinesisShardAssigner) {
+        this.kinesisShardAssigner = kinesisShardAssigner;
+        return this;
+    }
+
+    public KinesisStreamsSource<T> build() {
+        return new KinesisStreamsSource<>(
+                streamArn, consumerConfig, deserializationSchema, kinesisShardAssigner);
+    }
+}
--- a/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/KinesisStreamsSourceConfigUtil.java
+++ b/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/KinesisStreamsSourceConfigUtil.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.connector.kinesis.source.config;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.util.Preconditions;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Properties;
+
+import static org.apache.flink.connector.kinesis.source.config.KinesisStreamsSourceConfigConstants.DEFAULT_STREAM_TIMESTAMP_DATE_FORMAT;
+import static org.apache.flink.connector.kinesis.source.config.KinesisStreamsSourceConfigConstants.STREAM_INITIAL_TIMESTAMP;
+import static org.apache.flink.connector.kinesis.source.config.KinesisStreamsSourceConfigConstants.STREAM_TIMESTAMP_DATE_FORMAT;
+
+/** Utility functions to use with {@link KinesisStreamsSourceConfigConstants}. */
+@Internal
+public class KinesisStreamsSourceConfigUtil {
+
+    private KinesisStreamsSourceConfigUtil() {
+        // private constructor to prevent initialization of utility class.
+    }
+
+    /**
+     * Parses the timestamp in which to start consuming from the stream, from the given properties.
+     *
+     * @param consumerConfig the properties to parse timestamp from
+     * @return the timestamp
+     */
+    public static Date parseStreamTimestampStartingPosition(final Properties consumerConfig) {
+        Preconditions.checkNotNull(consumerConfig);
+        String timestamp = consumerConfig.getProperty(STREAM_INITIAL_TIMESTAMP);
+
+        try {
+            String format =
+                    consumerConfig.getProperty(
+                            STREAM_TIMESTAMP_DATE_FORMAT, DEFAULT_STREAM_TIMESTAMP_DATE_FORMAT);
+            SimpleDateFormat customDateFormat = new SimpleDateFormat(format);
+            return customDateFormat.parse(timestamp);
+        } catch (IllegalArgumentException | NullPointerException exception) {
+            throw new IllegalArgumentException(exception);
+        } catch (ParseException exception) {
+            return new Date((long) (Double.parseDouble(timestamp) * 1000));
+        }
+    }
+}
--- a/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
+++ b/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
@@ -1,0 +1,349 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.connector.kinesis.source.enumerator;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.api.connector.source.ReaderInfo;
+import org.apache.flink.api.connector.source.SplitEnumerator;
+import org.apache.flink.api.connector.source.SplitEnumeratorContext;
+import org.apache.flink.api.connector.source.SplitsAssignment;
+import org.apache.flink.connector.kinesis.source.config.KinesisStreamsSourceConfigConstants.InitialPosition;
+import org.apache.flink.connector.kinesis.source.exception.KinesisStreamsSourceException;
+import org.apache.flink.connector.kinesis.source.proxy.StreamProxy;
+import org.apache.flink.connector.kinesis.source.split.KinesisShardSplit;
+import org.apache.flink.connector.kinesis.source.split.StartingPosition;
+
+import org.apache.flink.shaded.guava30.com.google.common.collect.ImmutableList;
+import org.apache.flink.shaded.guava30.com.google.common.collect.ImmutableMap;
+import org.apache.flink.shaded.guava30.com.google.common.collect.ImmutableSet;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.services.kinesis.model.Shard;
+
+import javax.annotation.Nullable;
+
+import java.io.IOException;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Properties;
+import java.util.Set;
+
+import static org.apache.flink.connector.kinesis.source.config.KinesisStreamsSourceConfigConstants.DEFAULT_SHARD_DISCOVERY_INTERVAL_MILLIS;
+import static org.apache.flink.connector.kinesis.source.config.KinesisStreamsSourceConfigConstants.DEFAULT_STREAM_INITIAL_POSITION;
+import static org.apache.flink.connector.kinesis.source.config.KinesisStreamsSourceConfigConstants.SHARD_DISCOVERY_INTERVAL_MILLIS;
+import static org.apache.flink.connector.kinesis.source.config.KinesisStreamsSourceConfigConstants.STREAM_INITIAL_POSITION;
+import static org.apache.flink.connector.kinesis.source.config.KinesisStreamsSourceConfigUtil.parseStreamTimestampStartingPosition;
+
+/**
+ * This class is used to discover and assign Kinesis splits to subtasks on the Flink cluster. This
+ * runs on the JobManager.
+ */
+@Internal
+public class KinesisStreamsSourceEnumerator
+        implements SplitEnumerator<KinesisShardSplit, KinesisStreamsSourceEnumeratorState> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(KinesisStreamsSourceEnumerator.class);
+
+    private final SplitEnumeratorContext<KinesisShardSplit> context;
+    private final String streamArn;
+    private final Properties consumerConfig;
+    private final StreamProxy streamProxy;
+    private final KinesisShardAssigner shardAssigner;
+    private final ShardAssignerContext shardAssignerContext;
+
+    private final Map<Integer, Set<KinesisShardSplit>> splitAssignment = new HashMap<>();
+    private final Set<String> assignedSplitIds = new HashSet<>();
+    private final Set<KinesisShardSplit> unassignedSplits;
+
+    private String lastSeenShardId;
+
+    public KinesisStreamsSourceEnumerator(
+            SplitEnumeratorContext<KinesisShardSplit> context,
+            String streamArn,
+            Properties consumerConfig,
+            StreamProxy streamProxy,
+            KinesisShardAssigner shardAssigner,
+            KinesisStreamsSourceEnumeratorState state) {
+        this.context = context;
+        this.streamArn = streamArn;
+        this.consumerConfig = consumerConfig;
+        this.streamProxy = streamProxy;
+        this.shardAssigner = shardAssigner;
+        this.shardAssignerContext = new ShardAssignerContext(splitAssignment, context);
+        if (state == null) {
+            this.lastSeenShardId = null;
+            this.unassignedSplits = new HashSet<>();
+        } else {
+            this.lastSeenShardId = state.getLastSeenShardId();
+            this.unassignedSplits = state.getUnassignedSplits();
+        }
+    }
+
+    @Override
+    public void start() {
+        if (lastSeenShardId == null) {
+            context.callAsync(this::initialDiscoverSplits, this::assignSplits);
+        }
+
+        final long shardDiscoveryInterval =
+                Long.parseLong(
+                        consumerConfig.getProperty(
+                                SHARD_DISCOVERY_INTERVAL_MILLIS,
+                                String.valueOf(DEFAULT_SHARD_DISCOVERY_INTERVAL_MILLIS)));
+        context.callAsync(
+                this::periodicallyDiscoverSplits,
+                this::assignSplits,
+                shardDiscoveryInterval,
+                shardDiscoveryInterval);
+    }
+
+    @Override
+    public void handleSplitRequest(int subtaskId, @Nullable String requesterHostname) {
+        // Do nothing, since we assign splits eagerly
+    }
+
+    @Override
+    public void addSplitsBack(List<KinesisShardSplit> splits, int subtaskId) {
+        if (!splitAssignment.containsKey(subtaskId)) {
+            LOG.warn(
+                    "Unable to add splits back for subtask {} since it is not assigned any splits. Splits: {}",
+                    subtaskId,
+                    splits);
+            return;
+        }
+
+        for (KinesisShardSplit split : splits) {
+            splitAssignment.get(subtaskId).remove(split);
+            assignedSplitIds.remove(split.splitId());
+            unassignedSplits.add(split);
+        }
+
+        // Assign the unassignedSplits
+        // We did not discover any new splits, so we put in an empty list
+        assignSplits(Collections.emptyList(), null);
+    }
+
+    @Override
+    public void addReader(int subtaskId) {
+        splitAssignment.putIfAbsent(subtaskId, new HashSet<>());
+    }
+
+    @Override
+    public KinesisStreamsSourceEnumeratorState snapshotState(long checkpointId) throws Exception {
+        return new KinesisStreamsSourceEnumeratorState(unassignedSplits, lastSeenShardId);
+    }
+
+    @Override
+    public void close() throws IOException {
+        streamProxy.close();
+    }
+
+    private List<KinesisShardSplit> initialDiscoverSplits() {
+        List<Shard> shards = streamProxy.listShards(streamArn, lastSeenShardId);
+        return mapToSplits(
+                shards,
+                InitialPosition.valueOf(
+                        consumerConfig
+                                .getOrDefault(
+                                        STREAM_INITIAL_POSITION, DEFAULT_STREAM_INITIAL_POSITION)
+                                .toString()));
+    }
+
+    /**
+     * This method is used to discover Kinesis splits the job can subscribe to. It can be run in
+     * parallel, is important to not mutate any shared state.
+     *
+     * @return list of discovered splits
+     */
+    private List<KinesisShardSplit> periodicallyDiscoverSplits() {
+        List<Shard> shards = streamProxy.listShards(streamArn, lastSeenShardId);
+        // Any shard discovered after the initial startup should be read from the start, since they
+        // come from resharding
+        return mapToSplits(shards, InitialPosition.TRIM_HORIZON);
+    }
+
+    private List<KinesisShardSplit> mapToSplits(
+            List<Shard> shards, InitialPosition initialPosition) {
+        StartingPosition startingPosition;
+        switch (initialPosition) {
+            case LATEST:
+                // If LATEST is requested, we still set the starting position to the time of
+                // startup. This way, the job starts reading from a deterministic timestamp
+                // (i.e. time of job submission), even if it enters a restart loop immediately
+                // after submission.
+                startingPosition = StartingPosition.fromTimestamp(Instant.now());
+                break;
+            case AT_TIMESTAMP:
+                startingPosition =
+                        StartingPosition.fromTimestamp(
+                                parseStreamTimestampStartingPosition(consumerConfig).toInstant());
+                break;
+            case TRIM_HORIZON:
+            default:
+                startingPosition = StartingPosition.fromStart();
+        }
+
+        List<KinesisShardSplit> splits = new ArrayList<>();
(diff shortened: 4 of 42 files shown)


flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/SourceConfigConstants.java
```diff
@@ -0,0 +1,369 @@
...
+    public static final long DEFAULT_DEREGISTER_STREAM_CONSUMER_BACKOFF_BASE = 500L;
+
+    /** The maximum backoff time between each deregisterStreamConsumer attempt. */
+    public static final String DEREGISTER_STREAM_CONSUMER_BACKOFF_MAX =
+            "flink.stream.deregisterstreamconsumer.backoff.max";
+
+    public static final long DEFAULT_DEREGISTER_STREAM_CONSUMER_BACKOFF_MAX = 2000L;
+
+    /** The power constant for exponential backoff between each deregisterStreamConsumer attempt. */
+    public static final String DEREGISTER_STREAM_BACKOFF_EXPONENTIAL_CONSTANT =
```
darenwkt, line 225: `DEREGISTER_STREAM_BACKOFF_EXPONENTIAL_CONSTANT` -> `DEREGISTER_STREAM_CONSUMER_BACKOFF_EXPONENTIAL_CONSTANT`

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/PollingKinesisShardSplitReader.java
darenwkt, line 1: We should consider adding Test for this class

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/SourceConfigConstants.java
vahmed-hamdy: same

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/HashShardAssigner.java
```diff
@@ -0,0 +1,36 @@
...
+@PublicEvolving
+public class HashShardAssigner implements KinesisShardAssigner {
+    @Override
+    public int assign(KinesisShardSplit split, Context context) {
+        Integer[] availableSubtasks = new Integer[] {};
+        availableSubtasks = context.getRegisteredReaders().keySet().toArray(availableSubtasks);
+        if (availableSubtasks.length < 1) {
+            throw new IllegalArgumentException(
+                    "Expected at least one registered reader. Unable to assign split.");
+        }
```
z3d1k, line 33: nit: using `Preconditions.checkArgument` might be cleaner.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -0,0 +1,255 @@
...
+                    startingPosition.toString(),
+                    startingTimestamp));
+        }
+
+        return splits;
+    }
+
+    private void assignSplits(List<KinesisShardSplit> discoveredSplits, Throwable t) {
+        if (t != null) {
+            throw new FlinkRuntimeException("Failed to list shards.", t);
```
z3d1k, line 191: Can more specific exception be used here?

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/SourceConfigConstants.java
hlteoh37: addressed in previous comment

flink-connector-aws-kinesis-streams/src/test/java/org/apache/flink/connector/kinesis/source/examples/SourceFromKinesis.java
```diff
@@ -0,0 +1,55 @@
...
+ * KDS.
+ */
+public class SourceFromKinesis {
+
+    public static void main(String[] args) throws Exception {
+        final StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
+        env.enableCheckpointing(10_000);
+
+        Properties consumerConfig = new Properties();
+        consumerConfig.setProperty(AWSConfigConstants.AWS_REGION, "us-east-1");
```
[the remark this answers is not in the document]
hlteoh37: I am thinking of addressing this in followup PR. We want to unify the "AWS Client Config" setup. Same for "Apache HTTP Client config" setup

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
```diff
@@ -0,0 +1,148 @@
...
+        props.stringPropertyNames().forEach(key -> config.setString(key, props.getProperty(key)));
+        return config;
+    }
+
+    private KinesisStreamProxy createKinesisStreamProxy(Properties consumerConfig) {
+        SdkHttpClient httpClient =
+                AWSGeneralUtil.createSyncHttpClient(
+                        AttributeMap.builder().build(), ApacheHttpClient.builder());
+
+        AWSGeneralUtil.validateAwsCredentials(consumerConfig);
```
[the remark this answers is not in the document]
hlteoh37: Added, good shout

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/HashShardAssigner.java
hlteoh37: Done

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
hlteoh37: Great shout! Added a `KinesisStreamsSourceException`

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -0,0 +1,255 @@
...
+        }
+    }
+
+    @Override
+    public void start() {
+        if (lastSeenShardId == null) {
+            context.callAsync(this::initialDiscoverSplits, this::assignSplits);
+        }
+
+        context.callAsync(this::periodicallyDiscoverSplits, this::assignSplits, 10_000L, 10_000L);
```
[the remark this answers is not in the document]
hlteoh37: Yep - configured

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumeratorStateSerializer.java
```diff
@@ -0,0 +1,131 @@
...
+    @Override
+    public KinesisStreamsSourceEnumeratorState deserialize(
+            int version, byte[] serializedEnumeratorState) throws IOException {
+        try (ByteArrayInputStream bais = new ByteArrayInputStream(serializedEnumeratorState);
+                DataInputStream in = new DataInputStream(bais)) {
+
+            final int serializerVersion = in.readInt();
+
+            if (serializerVersion != getVersion()) {
+                throw new IOException("Trying to deserialize KinesisStreamsSourceEnumeratorState serialized with unsupported version of " + getVersion());
```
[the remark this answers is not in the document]
hlteoh37: Great shout. Changed the two version exceptions into `VersionMismatchException`, but others left as IOException because they are

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/split/KinesisShardSplitSerializer.java
```diff
@@ -0,0 +1,95 @@
...
+                out.writeBoolean(true);
+                out.writeObject(split.getStartingPosition().getStartingMarker());
+            }
+            out.flush();
+            return baos.toByteArray();
+        }
+    }
+
+    @Override
+    public KinesisShardSplit deserialize(int i, byte[] serialized) throws IOException {
```
z3d1k, line 69: Argument naming: `int i` does not provide any context on what this parameter is for. Matching name from interface will make this part more readable:
```
public KinesisShardSplit deserialize(int version, byte[] serialized) throws IOException
```

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -0,0 +1,321 @@
...
+    private final SplitEnumeratorContext<KinesisShardSplit> context;
+    private final String streamArn;
+    private final Properties consumerConfig;
+    private final StreamProxy streamProxy;
+    private final KinesisShardAssigner shardAssigner;
+    private final KinesisShardAssigner.Context shardAssignerContext;
+
+    private final Map<Integer, Set<KinesisShardSplit>> splitAssignment = new HashMap<>();
+    private final Set<String> assignedSplitIds = new HashSet<>();
+    private final Set<KinesisShardSplit> unassignedSplits;
```
z3d1k, line 76: `KinesisShardSplit` does not have overrides for `hashCode` and `equals` which may lead to unexpected behaviour.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/HashShardAssigner.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.flink.annotation.Experimental;
+import org.apache.flink.connector.kinesis.source.split.KinesisShardSplit;
+import org.apache.flink.util.Preconditions;
+
+/** An implementation of the {@link KinesisShardAssigner} that assigns splits by hashcode. */
+@Experimental
+public class HashShardAssigner implements KinesisShardAssigner {
+    @Override
+    public int assign(KinesisShardSplit split, Context context) {
+        Integer[] availableSubtasks = new Integer[] {};
```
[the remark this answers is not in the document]
hlteoh37: can't do a `toArray()` with primitive!

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/split/KinesisShardSplitSerializer.java
hlteoh37: Yep - done. TY

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/assigner/HashShardAssigner.java
```diff
@@ -0,0 +1,38 @@
...
+
+package org.apache.flink.connector.kinesis.source.enumerator.assigner;
+
+import org.apache.flink.annotation.Experimental;
+import org.apache.flink.connector.kinesis.source.enumerator.KinesisShardAssigner;
+import org.apache.flink.connector.kinesis.source.split.KinesisShardSplit;
+import org.apache.flink.util.Preconditions;
+
+/** An implementation of the {@link KinesisShardAssigner} that assigns splits by hashcode. */
+@Experimental
```
zentol, line 27: Consider not exposing these implementations directly, but hide them behind a factory method that only returns a `KinesisShardAssigner`.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/SourceConfigConstants.java
```diff
@@ -0,0 +1,371 @@
...
+         * Register the efo consumer eagerly for jobmanager to execute. De-register the efo consumer
+         * the same way as lazy does.
+         */
+        EAGER,
+        /** Do not register efo consumer programmatically. Do not de-register either. */
+        NONE
+    }
+
+    /** The RecordPublisher type (EFO|POLLING). */
+    public static final String RECORD_PUBLISHER_TYPE = "flink.stream.recordpublisher";
```
zentol, line 64: What are these constants / how are they used?

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/split/KinesisShardSplitSerializer.java
```diff
@@ -0,0 +1,93 @@
...
+                ObjectOutputStream out = new ObjectOutputStream(baos)) {
+
+            out.writeUTF(split.getStreamArn());
+            out.writeUTF(split.getShardId());
+            out.writeUTF(split.getStartingPosition().getShardIteratorType().toString());
+            if (split.getStartingPosition().getStartingMarker() == null) {
+                out.writeBoolean(false);
+            } else {
+                out.writeBoolean(true);
+                out.writeObject(split.getStartingPosition().getStartingMarker());
```
zentol, line 60: consider not using java serialization for this, unless you can guarantee that an `Instant` going through java serialization can be deserialized on a new JDK.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/SourceConfigConstants.java
hlteoh37: At the moment, they are not used. I've put them in here for now to track the currently available configurations in the existing KinesisConsumer and ensure we have parity. 

If desired, I can remove them, but I would prefer to keep them for now, and do a cleanup before we release (given that this is marked as @Experimental)

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/split/KinesisShardSplitSerializer.java
hlteoh37: This is a good catch. I will check the other Serializers too.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/assigner/HashShardAssigner.java
hlteoh37: Good suggestion. Done

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
```diff
@@ -0,0 +1,167 @@
...
+ * into the Flink job graph.
+ *
+ * <p>Exactly-once semantics. To leverage Flink's checkpointing mechanics for exactly-once stream
+ * processing, the Kinesis Source is implemented with the AWS Java SDK, instead of the officially
+ * recommended AWS Kinesis Client Library. The source will store its current progress in Flink
+ * checkpoint/savepoint, and will pick up from where it left off upon restore from the
+ * checkpoint/savepoint.
+ *
+ * <p>Initial starting points. The Kinesis Streams Source supports reads starting from TRIM_HORIZON,
+ * LATEST, and AT_TIMESTAMP.
```
usamj, line 70: Nit: "LATEST, **or** AT_TIMESTAMP"

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/split/StartingPosition.java
```diff
@@ -0,0 +1,94 @@
...
+    @Nullable
+    public Object getStartingMarker() {
+        return startingMarker;
+    }
+
+    public static StartingPosition fromTimestamp(final Instant timestamp) {
+        return new StartingPosition(AT_TIMESTAMP, timestamp);
+    }
+
+    public static StartingPosition continueFromSequenceNumber(final String sequenceNumber) {
```
dannycranmer, line 59: FYI, I spent some time looking through this so thought I would share my findings. I questioned whether we need to support the `restartFromSequenceNumber` like the legacy source, but I believe we do not and this is good.

dannycranmer (changes requested):

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
```diff
@@ -0,0 +1,167 @@
...
+ *
+ * @param <T> the data type emitted by the source
+ */
+@Experimental
+public class KinesisStreamsSource<T>
```
dannycranmer, line 75: Let's add a builder like the sink has

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/SourceConfigUtil.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Date;
+import java.util.Properties;
+
+import static org.apache.flink.connector.kinesis.source.config.SourceConfigConstants.DEFAULT_STREAM_TIMESTAMP_DATE_FORMAT;
+import static org.apache.flink.connector.kinesis.source.config.SourceConfigConstants.STREAM_INITIAL_TIMESTAMP;
+import static org.apache.flink.connector.kinesis.source.config.SourceConfigConstants.STREAM_TIMESTAMP_DATE_FORMAT;
+
+/** Utility functions to use with {@link SourceConfigConstants}. */
+@Internal
+public class SourceConfigUtil {
```
dannycranmer, line 34: nit: The name is pretty generic, can we rename to `KinesisStreamsSourceConfigUtil`. i realise this is implicit by the package, but my preference is to make the class name reflect the responsibility. Same for Constants

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/assigner/UniformShardAssigner.java
```diff
@@ -0,0 +1,53 @@
...
+            if (subtaskAssignmentSize < curMinAssignment) {
+                curMinAssignment = subtaskAssignmentSize;
+                selectedSubtask = subtaskId;
+            }
+        }
+
+        Preconditions.checkArgument(
+                selectedSubtask != -1,
+                "Expected at least one registered reader. Unable to assign split.");
+        return selectedSubtask;
```
dannycranmer, line 51: I ran this locally with 2 parallelism and 2 shards. Both shards were assigned to subTask 0 because `getCurrentSplitAssignment` was not updated in time, race condition.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/proxy/StreamProxy.java
```diff
@@ -0,0 +1,56 @@
...
+import software.amazon.awssdk.services.kinesis.model.GetRecordsResponse;
+import software.amazon.awssdk.services.kinesis.model.Shard;
+
+import javax.annotation.Nullable;
+
+import java.util.List;
+
+/** Interface for a StreamProxy to interact with Streams service in a given region. */
+@Internal
+public interface StreamProxy {
```
dannycranmer, line 33: You should add a `close()` method or make `Closable`

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    private KinesisStreamProxy createKinesisStreamProxy(Properties consumerConfig) {
+        SdkHttpClient httpClient =
+                AWSGeneralUtil.createSyncHttpClient(
+                        AttributeMap.builder().build(), ApacheHttpClient.builder());
+
+        AWSGeneralUtil.validateAwsCredentials(consumerConfig);
+        KinesisClient kinesisClient =
+                AWSClientUtil.createAwsSyncClient(
```
dannycranmer, line 159: These clients need to be closed, both the HTTP client and Kinesis client

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    @Override
+    public KinesisStreamsSourceEnumeratorState snapshotState(long checkpointId) throws Exception {
+        return new KinesisStreamsSourceEnumeratorState(
+                completedSplitIds, unassignedSplits, lastSeenShardId);
+    }
+
+    @Override
+    public void close() throws IOException {}
```
dannycranmer, line 157: @You need to close the Kinesis and HTTP client here; delegate to `StreamProxy::close`

dannycranmer (changes requested):

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
```diff
@@ -0,0 +1,167 @@
...
+    public SimpleVersionedSerializer<KinesisStreamsSourceEnumeratorState>
+            getEnumeratorCheckpointSerializer() {
+        return new KinesisStreamsSourceEnumeratorStateSerializer(new KinesisShardSplitSerializer());
+    }
+
+    private Configuration convertPropertiesToConfiguration(Properties props) {
+        Configuration config = new Configuration();
+        props.stringPropertyNames().forEach(key -> config.setString(key, props.getProperty(key)));
+        return config;
+    }
```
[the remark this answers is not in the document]
hlteoh37: you're right. Used a utility function instead

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
```diff
@@ -0,0 +1,167 @@
...
+public class KinesisStreamsSource<T>
+        implements Source<T, KinesisShardSplit, KinesisStreamsSourceEnumeratorState> {
+
+    private final String streamArn;
+    private final Properties consumerConfig;
+    private final DeserializationSchema<T> deserializationSchema;
+
+    public KinesisStreamsSource(
+            String streamArn,
+            Properties consumerConfig,
```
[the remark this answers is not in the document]
hlteoh37: Yes, that was the intention, but we needed to refactor aws-base to also handle configuration object rather than properties object. So thats why I created a followup here
https://issues.apache.org/jira/browse/FLINK-31990

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/PollingKinesisShardSplitReader.java
```diff
@@ -0,0 +1,152 @@
...
+    public void wakeUp() {
+        // Do nothing because we don't have any sleep mechanism
+    }
+
+    @Override
+    public void close() throws Exception {}
+
+    private static class KinesisRecordsWithSplitIds implements RecordsWithSplitIds<Record> {
+
+        private final Iterator<Record> recordsIterator;
```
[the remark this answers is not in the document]
hlteoh37: > We need to keep this as it might be a blocker for many users

Yes that is correct, we had a discussion recently, and we decided it's important. Created followup https://issues.apache.org/jira/browse/FLINK-32097

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/split/KinesisShardSplitState.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public KinesisShardSplit getKinesisShardSplit() {
+        return new KinesisShardSplit(
+                kinesisShardSplit.getStreamArn(),
+                kinesisShardSplit.getShardId(),
+                nextStartingPosition);
+    }
+
+    public String splitId() {
```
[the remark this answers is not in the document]
hlteoh37: Fixed

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/split/KinesisShardSplitState.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public String getShardId() {
+        return kinesisShardSplit.getShardId();
+    }
+
+    public StartingPosition getNextStartingPosition() {
+        return nextStartingPosition;
+    }
+
+    public void setNextStartingPosition(StartingPosition nextStartingPosition) {
```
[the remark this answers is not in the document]
hlteoh37: Good question, but yes, this class needs to be mutable because of the manner it's used in the `SourceReaderBase`. On assignment of split, it will be created, but on `snapshotState`, the state will be retrieved and we will save the `KinesisShardSplit` object returned into a savepoint. 

Because the reference of the `KinesisShardSplitState` is never updated in the `SourceReaderBase`, we need to make this mutable.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/split/StartingPosition.java
hlteoh37: My understanding was that `restartFromSequenceNumber` is only needed when we support aggregated records (then we need to read from `AT_SEQUENCE_NUMBER` rather than `AFTER_SEQUENCE_NUMBER`. So if we support KPL aggregation, we will need to support this. Does this match your understanding ?

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/SourceConfigUtil.java
hlteoh37: Ok, that makes sense! Adjusted.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/assigner/UniformShardAssigner.java
hlteoh37: That is correct. Thanks for finding the bug! Turns out it's because shards discovered in a batch does not consider the other shards already assigned within the same batch. I thought about providing a "batch assign" method in the `KinesisShardAssigner` interface, but thought that it would be a little extra, because we would have to first loop through to remove completed shards, then we would need to loop through again to assign them shards.

Do you think we should add a batch `assign` interface instead?

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
hlteoh37: Yes this is a good catch. wired in now.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
hlteoh37: This is correct. It's a pain to handle the HTTP client like this. Will look into passing the HTTP client builder into the AWS SDK builder instead. That way the lifecycle is coupled.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -0,0 +1,371 @@
...
+    private final String streamArn;
+    private final Properties consumerConfig;
+    private final StreamProxy streamProxy;
+    private final KinesisShardAssigner shardAssigner;
+    private final ShardAssignerContext shardAssignerContext;
+
+    private final Map<Integer, Set<KinesisShardSplit>> splitAssignment = new HashMap<>();
+    private final Set<String> assignedSplitIds = new HashSet<>();
+    private final Set<KinesisShardSplit> unassignedSplits;
+    private final Set<String> completedSplitIds;
```
PatrickRen, line 83: I'm not quite familiar with Kinesis so please correct me if I'm wrong. By `completed` does that mean there won't be new records in the shard (like a bounded shard), or the shard is going to be removed? 

The only usage of this `completedSplitIds` is to avoid assigning a completed shard to readers, so if the SDK can promise that we won't get completed shards using the last seen shard ID, it's not necessary to have the `completedSplitIds` I guess.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/PollingKinesisShardSplitReader.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+
+    @Override
+    public void wakeUp() {
+        // Do nothing because we don't have any sleep mechanism
+    }
+
+    @Override
+    public void close() throws Exception {}
```
PatrickRen, line 114: The Kinesis client should be closed here I think.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/SourceConfigConstants.java
PatrickRen, line 64: It's a bit weird that all configs start with `flink.`, which look like configs for Flink instead of the connector 🤔

PatrickRen (commented):
Thanks for the new source @hlteoh37 ! I left some comments. Please correct me if there's anything wrong as I'm not quite familiar with Kinesis.

I didn't look deeply into tests, but I can't find any integration tests (using the source in a real Flink job and run with mini cluster). It'll be nice to add some later.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/assigner/ShardAssignerFactory.java
```diff
@@ -0,0 +1,41 @@
...
+public class ShardAssignerFactory {
+    private ShardAssignerFactory() {
+        // No-op private constructor to prevent instantiation of class
+    }
+
+    public static KinesisShardAssigner uniformShardAssigner() {
+        return new UniformShardAssigner();
+    }
+
+    public static KinesisShardAssigner hashShardAssigner() {
```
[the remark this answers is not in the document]
hlteoh37: Ah, this was meant to be configurable. Will make it appear in the constructor.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -0,0 +1,371 @@
...
+                    "Unable to add splits back for subtask {} since it is not assigned any splits. Splits: {}",
+                    subtaskId,
+                    splits);
+            return;
+        }
+
+        for (KinesisShardSplit split : splits) {
+            splitAssignment.get(subtaskId).remove(split);
+            assignedSplitIds.remove(split.splitId());
+            unassignedSplits.add(split);
```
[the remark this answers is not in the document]
hlteoh37: Good catch. Added.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
hlteoh37: > By completed does that mean there won't be new records in the shard (like a bounded shard), or the shard is going to be removed?

When a shard is completed, no new records can be added to that shard, correct. The shard can still be read from. It will be deleted after all data in the shard exceeds the data retention period.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/PollingKinesisShardSplitReader.java
hlteoh37: Ah, yes. I made some changes. Let me push

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/SourceConfigConstants.java
hlteoh37: Thats a great callout. I had thought that was weird too... At the moment, we just copy from the previous connector. I wanted to rename these configs as part of this JIRA https://issues.apache.org/jira/browse/FLINK-31990

hlteoh37:
Thanks for the feedback @PatrickRen ! I've created a followup subtask for IT tests, because otherwise this PR will be super long 
https://issues.apache.org/jira/browse/FLINK-31986

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
hlteoh37: Ok added.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Override
+    public SourceReader<T, KinesisShardSplit> createReader(SourceReaderContext readerContext)
+            throws Exception {
+        FutureCompletingBlockingQueue<RecordsWithSplitIds<Record>> elementsQueue =
+                new FutureCompletingBlockingQueue<>();
+        KinesisStreamProxy kinesisStreamProxy = createKinesisStreamProxy(consumerConfig);
+        Supplier<PollingKinesisShardSplitReader> splitReaderSupplier =
+                () -> new PollingKinesisShardSplitReader(kinesisStreamProxy);
+        KinesisStreamsRecordEmitter<T> recordEmitter =
```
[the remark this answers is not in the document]
hlteoh37: Great spot. Added

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -0,0 +1,338 @@
...
+     *
+     * @param discoveredSplits list of discovered splits
+     * @param t throwable thrown when discovering splits. Will be null if no throwable thrown.
+     */
+    private void assignSplits(List<KinesisShardSplit> discoveredSplits, Throwable t) {
+        if (t != null) {
+            throw new KinesisStreamsSourceException("Failed to list shards.", t);
+        }
+
+        if (context.registeredReaders().isEmpty()) {
```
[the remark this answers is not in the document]
hlteoh37: Hm, to be honest I don't expect this to be ever hit in our scenario, but I put this in because the lifecycle of `listShard()` is on JM, and lifecycle of `registerReader` is on TM, so we don't have a strong guarantee that we will always registerReader before listShard. 

Also, on partial failover, the reader that experienced failover is removed.  If we support partial failover, and all subtasks fail partially, we could end up with no registered readers.


In this case, since we can put it into unassignedSplits, we should be ok to handle this transparently.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/assigner/HashShardAssigner.java
```diff
@@ -0,0 +1,38 @@
...
+package org.apache.flink.connector.kinesis.source.enumerator.assigner;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.connector.kinesis.source.enumerator.KinesisShardAssigner;
+import org.apache.flink.connector.kinesis.source.split.KinesisShardSplit;
+import org.apache.flink.util.Preconditions;
+
+/** An implementation of the {@link KinesisShardAssigner} that assigns splits by hashcode. */
+@Internal
+public class HashShardAssigner implements KinesisShardAssigner {
```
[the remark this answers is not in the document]
hlteoh37: Ok. will delete

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/assigner/ShardAssignerFactory.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.apache.flink.annotation.Experimental;
+import org.apache.flink.connector.kinesis.source.enumerator.KinesisShardAssigner;
+
+/**
+ * Factory that provides an instance of {@link KinesisShardAssigner} pre-packaged with the
+ * connector.
+ */
+@Experimental
+public class ShardAssignerFactory {
```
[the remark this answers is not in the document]
hlteoh37: We implemented this to make `UniformShardAssigner` `@Internal` ( presumably so we can change the implementation logic)

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -0,0 +1,338 @@
...
+            assignSplitToSubtask(split, newSplitAssignments);
+        }
+        unassignedSplits.clear();
+        for (KinesisShardSplit split : discoveredSplits) {
+            assignSplitToSubtask(split, newSplitAssignments);
+        }
+
+        updateLastSeenShardId(discoveredSplits);
+        updateSplitAssignment(newSplitAssignments);
+        context.assignSplits(new SplitsAssignment<>(newSplitAssignments));
```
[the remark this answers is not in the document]
hlteoh37: Great callout. 

So I have checked that should be effectively synchronous across components due to our RPC setup.


1. **On JM:** assignSplit() runs in the coordinatorThread. [code here](https://github.com/apache/flink/blob/master/flink-runtime/src/main/java/org/apache/flink/runtime/source/coordinator/SourceCoordinatorContext.java#L323). (This is the same thread that handles checkpointCoordinator() that triggers checkpoint on SourceCoordinator). This means the order of calls will be maintained here. 
2. In both cases, these make RPC calls to the TM. [assignSplit() calls `sendOperatorEventToTask()`](https://github.com/apache/flink/blob/2c25292272306a6eb094be40be0a0fa5fc3cf644/flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutorGateway.java#L284) and `checkpointCoordinator()` [calls `triggerCheckpoint()`](https://github.com/apache/flink/blob/2c25292272306a6eb094be40be0a0fa5fc3cf644/flink-runtime/src/main/java/org/apache/flink/runtime/taskexecutor/TaskExecutorGateway.java#L141). 
3. Since RPC calls are FIFO, we can be assured that the `assignSplit()` will be always processed before `checkpointCoordinator()`

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/KinesisStreamsRecordEmitter.java
```diff
@@ -0,0 +1,56 @@
...
+    public KinesisStreamsRecordEmitter(DeserializationSchema<T> deserializationSchema) {
+        this.deserializationSchema = deserializationSchema;
+    }
+
+    @Override
+    public void emitRecord(
+            Record element, SourceOutput<T> output, KinesisShardSplitState splitState)
+            throws Exception {
+        output.collect(
+                deserializationSchema.deserialize(element.data().asByteArray()),
```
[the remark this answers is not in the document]
hlteoh37: Ok, implemented

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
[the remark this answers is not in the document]
dannycranmer: Thanks for the deep dive @hlteoh37 !

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -0,0 +1,347 @@
...
+            return;
+        }
+
+        for (KinesisShardSplit split : splits) {
+            splitAssignment.get(subtaskId).remove(split);
+            assignedSplitIds.remove(split.splitId());
+            unassignedSplits.add(split);
+        }
+
+        // Assign the newly returned splits
```
dannycranmer, line 144: Can you elaborate on this comment as the `emptyList()` here is confusing based on the comment.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -0,0 +1,347 @@
...
+        return new KinesisStreamsSourceEnumeratorState(unassignedSplits, lastSeenShardId);
+    }
+
+    @Override
+    public void close() throws IOException {
+        streamProxy.close();
+    }
+
+    private List<KinesisShardSplit> initialDiscoverSplits() {
+        LOG.info("initialDiscoverSplits");
```
dannycranmer, line 164: Either remove this or add more details, for example the Stream ARN. Looks like a debug comment

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -0,0 +1,347 @@
...
+    /**
+     * This method assigns a given set of Kinesis splits to the readers currently registered on the
+     * cluster. This assignment is done via a side-effect on the {@link SplitEnumeratorContext}
+     * object.
+     *
+     * @param discoveredSplits list of discovered splits
+     * @param t throwable thrown when discovering splits. Will be null if no throwable thrown.
+     */
+    private void assignSplits(List<KinesisShardSplit> discoveredSplits, Throwable t) {
+        LOG.info("assignSplits");
```
dannycranmer, line 226: Same as above, looks like a debugging comment

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -0,0 +1,347 @@
...
+
+    /**
+     * This method assigns a given set of Kinesis splits to the readers currently registered on the
+     * cluster. This assignment is done via a side-effect on the {@link SplitEnumeratorContext}
+     * object.
+     *
+     * @param discoveredSplits list of discovered splits
+     * @param t throwable thrown when discovering splits. Will be null if no throwable thrown.
+     */
+    private void assignSplits(List<KinesisShardSplit> discoveredSplits, Throwable t) {
```
dannycranmer, line 225: nit: Instead of `t` use a meaningful arg name

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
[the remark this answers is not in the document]
dannycranmer: Following on from this, it seems if there is at least 1 reader we will blindly assign splits to it based on the shard assigner output. I wonder if we should skip assignment when the reader is not registered? Another rejected option is to assign all splits to the registered readers, but this will result in skew so not a good idea.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/assigner/UniformShardAssigner.java
```diff
@@ -0,0 +1,59 @@
...
+            int subtaskAssignmentSize =
+                    splitAssignment.getOrDefault(subtaskId, Collections.emptySet()).size()
+                            + pendingSplitAssignments
+                                    .getOrDefault(subtaskId, Collections.emptyList())
+                                    .size();
+            if (subtaskAssignmentSize < curMinAssignment) {
+                curMinAssignment = subtaskAssignmentSize;
+                selectedSubtask = subtaskId;
+            }
+        }
```
dannycranmer, line 52: I commented on this above. If for whatever reason only 1 reader is registered on the first listShards we would end up assigning ALL shards to a single subtask. I think instead we should wait for the readers to attach since we assume that all subtasks will attach, right?

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/PollingKinesisShardSplitReader.java
dannycranmer: Resolved but not updated, unresolved

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/serialization/KinesisDeserializationSchemaWrapper.java
```diff
@@ -0,0 +1,60 @@
...
+        this.deserializationSchema = deserializationSchema;
+    }
+
+    @Override
+    public void open(DeserializationSchema.InitializationContext context) throws Exception {
+        this.deserializationSchema.open(context);
+    }
+
+    @Override
+    public void deserialize(Record record, String stream, String shardId, Collector<T> output)
```
dannycranmer, line 51: The old version supplies additional parameters, why did we drop them?

https://github.com/apache/flink-connector-aws/blob/main/flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/serialization/KinesisDeserializationSchema.java#L64

dannycranmer (changes requested):

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/KinesisStreamsSourceConfigUtil.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Date;
+import java.util.Properties;
+
+import static org.apache.flink.connector.kinesis.source.config.KinesisStreamsSourceConfigConstants.DEFAULT_STREAM_TIMESTAMP_DATE_FORMAT;
+import static org.apache.flink.connector.kinesis.source.config.KinesisStreamsSourceConfigConstants.STREAM_INITIAL_TIMESTAMP;
+import static org.apache.flink.connector.kinesis.source.config.KinesisStreamsSourceConfigConstants.STREAM_TIMESTAMP_DATE_FORMAT;
+
+/** Utility functions to use with {@link KinesisStreamsSourceConfigConstants}. */
+@Internal
+public class KinesisStreamsSourceConfigUtil {
```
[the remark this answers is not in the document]
hlteoh37: Yes. I have added this to https://issues.apache.org/jira/browse/FLINK-31990

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSourceBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+ * <ul>
+ *   <li>{@code kinesisShardAssigner} will be {@link UniformShardAssigner}
+ * </ul>
+ *
+ * @param <T> type of elements that should be read from the source stream
+ */
+@Experimental
+public class KinesisStreamsSourceBuilder<T> {
+    private String streamArn;
+    private Properties consumerConfig = new Properties();
```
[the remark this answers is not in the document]
hlteoh37: Ok done. We can add feature to parse region from the ARN in a followup

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/KinesisStreamsSourceConfigUtil.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * Parses the timestamp in which to start consuming from the stream, from the given properties.
+     *
+     * @param consumerConfig the properties to parse timestamp from
+     * @return the timestamp
+     */
+    public static Date parseStreamTimestampStartingPosition(final Properties consumerConfig) {
+        String timestamp = consumerConfig.getProperty(STREAM_INITIAL_TIMESTAMP);
```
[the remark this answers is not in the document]
hlteoh37: done

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -0,0 +1,347 @@
...
+        if (lastSeenShardId == null) {
+            context.callAsync(this::initialDiscoverSplits, this::assignSplits);
+        }
+
+        final long shardDiscoveryInterval =
+                Long.parseLong(
+                        consumerConfig.getProperty(
+                                SHARD_DISCOVERY_INTERVAL_MILLIS,
+                                String.valueOf(DEFAULT_SHARD_DISCOVERY_INTERVAL_MILLIS)));
+        context.callAsync(
```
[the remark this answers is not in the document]
hlteoh37: > This will result in scheduling 2 in parallel initialDiscoverSplits and periodicallyDiscoverSplits . I wonder if we should instead setup periodicallyDiscoverSplits from the end of initialDiscoverSplits method in the lastSeenShardId is null case.

We set an initial delay for the periodic scheduling of `shardDiscoveryInterval`.

There is no race condition that wouldn't already exist on periodic call.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
hlteoh37: Ok done.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
hlteoh37: removed

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
hlteoh37: removed

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
hlteoh37: done.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -0,0 +1,347 @@
...
+        private ShardAssignerContext(
+                Map<Integer, Set<KinesisShardSplit>> splitAssignment,
+                SplitEnumeratorContext<KinesisShardSplit> splitEnumeratorContext) {
+            this.splitAssignment = splitAssignment;
+            this.splitEnumeratorContext = splitEnumeratorContext;
+        }
+
+        private ShardAssignerContext withPendingSplitAssignments(
+                Map<Integer, List<KinesisShardSplit>> pendingSplitAssignments) {
+            ImmutableMap.Builder<Integer, List<KinesisShardSplit>> mapBuilder =
```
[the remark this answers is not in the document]
hlteoh37: We don't want the user thinking that they can assign splits from the `ShardAssignerContext`.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/assigner/UniformShardAssigner.java
hlteoh37: Ok added.

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/serialization/KinesisDeserializationSchemaWrapper.java
hlteoh37: They are part of the `Record` datatype

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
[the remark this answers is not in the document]
hlteoh37: Added a check to only assign splits when there are enough readers

hlteoh37:
@dannycranmer @PatrickRen. This PR is ready for final review. Let me know if you have any further comments

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
```diff
@@ -0,0 +1,204 @@
...
+    private final Properties consumerConfig;
+    private final KinesisDeserializationSchema<T> deserializationSchema;
+    private final KinesisShardAssigner kinesisShardAssigner;
+
+    KinesisStreamsSource(
+            String streamArn,
+            Properties consumerConfig,
+            KinesisDeserializationSchema<T> deserializationSchema,
+            KinesisShardAssigner kinesisShardAssigner) {
+        Preconditions.checkNotNull(streamArn);
```
usamj, line 93: Do we want to do a notEmpty check on the stream ARN?

hlteoh37: sure

PatrickRen (approved):
@hlteoh37 Thanks for the update! LGTM. 

Also it'll be better to register more Kinesis-specific metrics, which will help users to monitor or debug their Flink job (it's totally optional for this PR). 

I'll give a green light for the FLIP-27 related part, and for Kinesis-specific logics maybe @dannycranmer could make another confirmation 😄

hlteoh37:
Thanks @PatrickRen. Thats a great callout. I've created a separate subtask to track that, as well as logging. https://issues.apache.org/jira/browse/FLINK-32229

flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSourceBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * <p>The following example shows the minimum setup to create a {@link KinesisStreamsSource} that
+ * reads String values from a Kinesis Data Streams stream with ARN of
+ * arn:aws:kinesis:us-east-1:012345678901:stream/your_stream_name.
+ *
+ * <pre>{@code
+ * KinesisStreamsSource<String> kdsSource =
+ *                 KinesisStreamsSource.<String>builder()
+ *                         .setStreamArn("arn:aws:kinesis:us-east-1:012345678901:stream/your_stream_name")
+ *                         .setDeserializationSchema(new SimpleStringSchema())
```
dannycranmer, line 41: I think properties (`consumerConfig`) are required too

dannycranmer (approved):
LGTM, thanks @hlteoh37

PR #77, opened 2023-06-12
[FLINK-31990][Connectors/Kinesis] Use Configuration object instead of…

hlteoh37:
… properties in KDS Source

## Purpose of the change

Use `Configuration` object for specifying Source config instead of `Properties`.

This makes it easier to view property description and default values in one place. Also standardises the configuration for Table API usages.

## Verifying this change

## Significant changes


vahmed-hamdy (approved):
LGTM! thanks @hlteoh37

PR #83, opened 2023-06-17
[FLINK-32372] Push PR support build and test with recently release fl…

Samrat002:
…ink version 1.16.2 and 1.17.1

## Purpose of the change

Support build and test for flink 1.16.2 and 1.17.1 flink release.

## Significant changes

### DIFF
--- a/.github/workflows/push_pr.yml
+++ b/.github/workflows/push_pr.yml
@@ -26,7 +26,7 @@ jobs:
     uses: ./.github/workflows/common.yml
     strategy:
       matrix:
-        flink: [1.16.1, 1.17.0]
+        flink: [1.16.2, 1.17.1]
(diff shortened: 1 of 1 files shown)


.github/workflows/push_pr.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     uses: ./.github/workflows/common.yml
     strategy:
       matrix:
-        flink: [1.16.1, 1.17.0]
+        flink: [1.16.1, 1.16.2, 1.17.0, 1.17.1]
```
hlteoh37, line 29: We should just upgrade the current versions, We shouldn't need to do 1.16.1 and 1.17.0!

reswqa (approved):

PR #92, opened 2023-08-28
[hotfix] Add MiniClusterExtension to ITCase tests

hlteoh37:
## Purpose of the change

* [Text for ITCase MiniCluster ArchUnitTest has been changed for Flink 1.18](https://github.com/apache/flink/pull/22399). This PR removes the need for this exception violation in ITCase, by registering a local MiniCluster environment in the Firehose ITCase.
* Change in the github workflow is merely to show that this case passes against `1.18-SNAPSHOT`. It will be removed before merging.

## Verifying this change
This change is to fix breaking ArchUnitTests.

## Significant changes

### DIFF
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
@@ -30,10 +30,6 @@ import org.apache.flink.connector.kinesis.source.proxy.StreamProxy;
 import org.apache.flink.connector.kinesis.source.split.KinesisShardSplit;
 import org.apache.flink.connector.kinesis.source.split.StartingPosition;
 
-import org.apache.flink.shaded.guava30.com.google.common.collect.ImmutableList;
-import org.apache.flink.shaded.guava30.com.google.common.collect.ImmutableMap;
-import org.apache.flink.shaded.guava30.com.google.common.collect.ImmutableSet;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.services.kinesis.model.Shard;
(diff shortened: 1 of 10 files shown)


hlteoh37:
Verified that checks against 1.18-SNAPSHOT pass
image

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
```diff
@@ -30,10 +30,6 @@
 import org.apache.flink.connector.kinesis.source.split.KinesisShardSplit;
 import org.apache.flink.connector.kinesis.source.split.StartingPosition;
 
-import org.apache.flink.shaded.guava30.com.google.common.collect.ImmutableList;
```
Samrat002, line 33: Are we getting rid of guava ? 
I dont see any dependency removal of guava in pom file ?

Samrat002 (commented):
com.google.common.collect.ImmutableSet and java.util.stream.Stream are two different concepts in Java that serve distinct purposes:

*com.google.common.collect.ImmutableSet:*

ImmutableSet is part of the Google Guava library, which provides various utility classes and data structures. ImmutableSet is used to create a set (a collection that doesn't allow duplicate elements) that cannot be modified after creation. This is particularly useful in scenarios where you want to ensure that the contents of a set remain constant throughout its lifecycle. Once an ImmutableSet is created, you can't add, remove, or modify elements in it.

For example:

```
ImmutableSet<String> colors = ImmutableSet.of("red", "green", "blue");
```

*java.util.stream.Stream:*
Stream is part of the Java Streams API introduced in Java 8. A Stream represents a sequence of elements that can be processed in various ways, such as filtering, mapping, and reducing. Streams allow you to perform operations on a collection of data in a functional and often more concise manner compared to traditional loops. Unlike ImmutableSet, a Stream doesn't hold data itself; it's a pipeline through which data can flow.

For example, you can use a Stream to filter and transform a list of numbers:

```
List<Integer> numbers = Arrays.asList(1, 2, 3, 4, 5);
Stream<Integer> squaredEvenNumbers = numbers.stream()
    .filter(n -> n % 2 == 0)
    .map(n -> n * n);
```

In summary, ImmutableSet is a specific data structure that provides an immutable set of elements, while Stream is a broader concept that represents a sequence of elements and allows you to perform various operations on that sequence. The choice between them depends on the use case: if you need an immutable set, you'd go for ImmutableSet, and if you need to process and manipulate data in a functional way, you'd use Stream.

Some Code changes here in test cases changes from ImmutableMap/ImmutableSet to stream then wrapping it up with `Collections.unmodifiableMap(`. Isn't that adds extra computations ?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/enumerator/KinesisStreamsSourceEnumerator.java
hlteoh37: This comes from the shaded guava! We are removing this dependency so that we don't have to keep upgrading this when Flink upgrades shaded guava

hlteoh37: All the changes that changes `ImmutableXXX` to `Streams` were in tests. We do not need to use a `ImmutableXXX`

Samrat002 (approved):
Thank you for thorough explanation, @hlteoh37 , regarding the core modifications wrt wrapping using Collections + performance and their anticipated minimal/no impact.

I'm in agreement, the changes looks good to me 🚀.

PR #93, opened 2023-09-08
[FLINK-24943][Connectors / Kinesis] Explicitly create KryoSerializer for SequenceNumber class in Kinesis Consumer

Cyberness:
## Purpose of the change

Create KryoSerializer for SequenceNumber class explicitly to allow disabling of Generic Types via disableGenericTypes() call.

## Brief change log

Create KinesisStateUtil class with static method createShardsStateSerializer that explicitly assigns KryoSerializer to SequenceNumber class and use this method to initialize state
Create unit test that checks for compatibility of previous TypeInformation based serializer and explicitly created KryoSerializer
Replace Mokito calls with utility MockStreamingRuntimeContext to improve testability and unify initialization method of Streaming Runtime Context

## Verifying this change

- *Added unit tests*

## Significant changes


dannycranmer (approved):
LGTM, thanks @Cyberness

PR #105, opened 2023-10-15
[FLINK-33181] Allow a table definition can be used to read & write data to.

vtkhanh:
## Purpose of the change

Allow a table can be used as a source and sink when it's defined with both consumer & producer options. The ability is achieved by not verifying an option if it's *indicated* as consumer (by `scan.` prefix) or producer (by `sink.` prefix) option when creating a source/sink based on the table definition.

Example: following table definition can be used for both source & sink operators:

```
%flink.ssql(type=update)

DROP TABLE IF EXISTS `kds_input`;
CREATE TABLE `kds_input` (
`some_string` STRING,
`some_int` BIGINT,
`time` AS PROCTIME()
) WITH (
'connector' = 'kinesis',
'stream' = 'ExampleInputStream',
'aws.region' = 'us-east-1',
'scan.stream.initpos' = 'LATEST',
'sink.fail-on-error' = 'true',
'format' = 'csv'
);
```

## Verifying this change

- *Added unit tests*

## Significant changes

### DIFF
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/table/util/KinesisStreamsConnectorOptionsUtils.java
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/table/util/KinesisStreamsConnectorOptionsUtils.java
@@ -62,6 +62,8 @@ public class KinesisStreamsConnectorOptionsUtils {
     /** Key for accessing kinesisAsyncClient properties. */
     public static final String KINESIS_CLIENT_PROPERTIES_KEY = "sink.client.properties";
 
+    public static final String CONSUMER_PREFIX = "scan.";
+
     private final AsyncClientOptionsUtils asyncClientOptionsUtils;
     private final AsyncSinkConfigurationValidator asyncSinkconfigurationValidator;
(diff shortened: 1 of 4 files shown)


vahmed-hamdy:
Hi  @vtkhanh, Could you please describe the use case more. It feels like an anti-pattern to use the same stream as source and sink. Kinesis Table API source and sink implementations were intentionally separated post 1.15.

dannycranmer:
>  Kinesis Table API source and sink implementations were intentionally separated post 1.15.

No we kept them together, and supported the legacy configs for backwards compatibility.

dannycranmer (approved):
LGTM, thanks @vtkhanh

dannycranmer:
Actions are failing due to a recent bug, will fix

vtkhanh:
> Hi @vtkhanh, Could you please describe the use case more. It feels like an anti-pattern to use the same stream as source and sink. Kinesis Table API source and sink implementations were intentionally separated post 1.15.

I dont have a concrete usecase in production which a stream is used for both read from and write to within a Flink job. But we can find it useful in testing when we want to populate random data into a stream, and then read from it, by using one table definition, instead of 2 separate ones.

vtkhanh:
> Actions are failing due to a recent bug, will fix

Merged your fix and updated the PR. Thanks for looking @dannycranmer.

PR #128, opened 2024-02-09
[FLINK-34407][Connectors/Kinesis] Fix unstable test

z3d1k:
## Purpose of the change

Tests that use `FakeKinesisFanOutBehavioursFactory` occasionally run into issue while creating mock for Kinesis shard subscription:
```
java.lang.ClassCastException: org.mockito.codegen.Subscription$MockitoMock$1959866534 cannot be cast to org.reactivestreams.Subscription
```
This failure leads to `TimeoutException` in `FanOutShardSubscriber`, resulting in workflow timing out due to continuous retries.

Using fake subscription implementation instead of Mockito mock to resolve this issue.

## Verifying this change

This change is already covered by existing tests and can be verified by running project build in Github Actions.

## Significant changes


hlteoh37:
LGTM thanks @z3d1k

PR #129, opened 2024-02-12
[FLINK-32964][Connectors/AWS] Ensure that new instance of DefaultCredentialsProvider is created for each client

z3d1k:
## Purpose of the change

Explicitly use `.builder().build()` instead of `.create()` method while constructing DefaultCredentialsProvider.
This way new instance of credentials provider will be created for each AWS SDK client to prevent failures when one of the clients closes a singleton instance, resulting in failures in others.

## Verifying this change

- *Added unit test*

## Significant changes


hlteoh37 (approved):
LGTM thanks @z3d1k

hlteoh37 (approved):
LGTM. Thanks @z3d1k

PR #136, opened 2024-04-11
[FLINK-35022][Connector/DynamoDB] Add TypeInformed DDB Element Converter as default element converter

vahmed-hamdy:
## Purpose of the change

Add `DynamoDbTypeInformedElementConverter` to convert Elements to dynamoDb Sink using its provided type Info.

## Verifying this change

- Added unit tests

## Significant changes

### DIFF
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
@@ -1,0 +1,380 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.connector.dynamodb.sink;
+
+import org.apache.flink.annotation.PublicEvolving;
+import org.apache.flink.api.common.typeinfo.BasicArrayTypeInfo;
+import org.apache.flink.api.common.typeinfo.BasicTypeInfo;
+import org.apache.flink.api.common.typeinfo.NumericTypeInfo;
+import org.apache.flink.api.common.typeinfo.PrimitiveArrayTypeInfo;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.common.typeutils.CompositeType;
+import org.apache.flink.api.connector.sink2.SinkWriter;
+import org.apache.flink.api.java.tuple.Tuple;
+import org.apache.flink.api.java.typeutils.ObjectArrayTypeInfo;
+import org.apache.flink.api.java.typeutils.PojoTypeInfo;
+import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.api.java.typeutils.TupleTypeInfo;
+import org.apache.flink.connector.base.sink.writer.ElementConverter;
+import org.apache.flink.connector.dynamodb.table.converter.ArrayAttributeConverterProvider;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.FlinkRuntimeException;
+import org.apache.flink.util.Preconditions;
+
+import software.amazon.awssdk.core.SdkBytes;
+import software.amazon.awssdk.enhanced.dynamodb.AttributeConverter;
+import software.amazon.awssdk.enhanced.dynamodb.AttributeConverterProvider;
+import software.amazon.awssdk.enhanced.dynamodb.AttributeValueType;
+import software.amazon.awssdk.enhanced.dynamodb.EnhancedType;
+import software.amazon.awssdk.enhanced.dynamodb.TableSchema;
+import software.amazon.awssdk.enhanced.dynamodb.internal.mapper.BeanAttributeGetter;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.StaticTableSchema;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+
+import java.beans.BeanInfo;
+import java.beans.IntrospectionException;
+import java.beans.Introspector;
+import java.beans.PropertyDescriptor;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Function;
+
+/**
+ * A {@link ElementConverter} that converts an element to a {@link DynamoDbWriteRequest} using
+ * TypeInformation provided.
+ */
+@PublicEvolving
+public class DynamoDbTypeInformedElementConverter<T>
+        implements ElementConverter<T, DynamoDbWriteRequest> {
+
+    private final CompositeType<T> typeInfo;
+    private final boolean ignoreNulls;
+    private final TableSchema<T> tableSchema;
+
+    /**
+     * Creates a {@link DynamoDbTypeInformedElementConverter} that converts an element to a {@link
+     * DynamoDbWriteRequest} using the provided {@link CompositeType}. Usage: {@code new
+     * DynamoDbTypeInformedElementConverter<>(TypeInformation.of(MyPojoClass.class))}
+     *
+     * @param typeInfo The {@link CompositeType} that provides the type information for the element.
+     */
+    public DynamoDbTypeInformedElementConverter(CompositeType<T> typeInfo) {
+        this(typeInfo, true);
+    }
+
+    public DynamoDbTypeInformedElementConverter(CompositeType<T> typeInfo, boolean ignoreNulls) {
+
+        try {
+            this.typeInfo = typeInfo;
+            this.ignoreNulls = ignoreNulls;
+            this.tableSchema = createTableSchema(typeInfo);
+        } catch (IntrospectionException | IllegalStateException | IllegalArgumentException e) {
+            throw new FlinkRuntimeException("Failed to extract DynamoDb table schema", e);
+        }
+    }
+
+    @Override
+    public DynamoDbWriteRequest apply(T input, SinkWriter.Context context) {
+        Preconditions.checkNotNull(tableSchema, "TableSchema is not initialized");
+        try {
+            return DynamoDbWriteRequest.builder()
+                    .setType(DynamoDbWriteRequestType.PUT)
+                    .setItem(tableSchema.itemToMap(input, ignoreNulls))
+                    .build();
+        } catch (ClassCastException | IllegalArgumentException e) {
+            throw new FlinkRuntimeException(
+                    String.format(
+                            "Failed to convert %s to DynamoDbWriteRequest using %s",
+                            input, typeInfo),
+                    e);
+        }
+    }
+
+    private <AttributeT> TableSchema<AttributeT> createTableSchema(
+            CompositeType<AttributeT> typeInfo) throws IntrospectionException {
+        if (typeInfo instanceof RowTypeInfo) {
+            return (TableSchema<AttributeT>) createTableSchemaFromRowType((RowTypeInfo) typeInfo);
+        } else if (typeInfo instanceof PojoTypeInfo<?>) {
+            return createTableSchemaFromPojo((PojoTypeInfo<AttributeT>) typeInfo);
+        } else if (typeInfo instanceof TupleTypeInfo<?>) {
+            return createTableSchemaFromTuple((TupleTypeInfo<?>) typeInfo);
+        } else {
+            throw new IllegalArgumentException(String.format("Unsupported TypeInfo %s", typeInfo));
+        }
+    }
+
+    private TableSchema<Row> createTableSchemaFromRowType(RowTypeInfo typeInfo) {
+        StaticTableSchema.Builder<Row> tableSchemaBuilder =
+                StaticTableSchema.builder(typeInfo.getTypeClass());
+
+        String[] fieldNames = typeInfo.getFieldNames();
+        for (int i = 0; i < typeInfo.getArity(); i++) {
+            TypeInformation<?> fieldType = typeInfo.getTypeAt(i);
+            int finalI = i;
+            addAttribute(
+                    tableSchemaBuilder,
+                    fieldNames[finalI],
+                    (AttributeT) -> AttributeT.getField(finalI),
+                    (TypeInformation<? super Object>) fieldType);
+        }
+
+        return tableSchemaBuilder.build();
+    }
+
+    private <AttributeT> TableSchema<AttributeT> createTableSchemaFromTuple(
+            TupleTypeInfo<?> typeInfo) {
+        TypeInformation<?>[] fieldTypes = typeInfo.getFieldTypes();
+        String[] fieldNames = typeInfo.getFieldNames();
+
+        StaticTableSchema.Builder<?> tableSchemaBuilder =
+                StaticTableSchema.builder(typeInfo.getTypeClass());
+        for (int i = 0; i < fieldNames.length; i++) {
+            int finalI = i;
+            addAttribute(
+                    tableSchemaBuilder,
+                    fieldNames[finalI],
+                    (tuple) -> ((Tuple) tuple).getField(finalI),
+                    (TypeInformation<?>) fieldTypes[i]);
+        }
+
+        return (TableSchema<AttributeT>) tableSchemaBuilder.build();
+    }
+
+    private <AttributeT> TableSchema<AttributeT> createTableSchemaFromPojo(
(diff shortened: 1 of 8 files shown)


flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
```diff
@@ -0,0 +1,245 @@
...
+            }
+        }
+
+        return map;
+    }
+
+    private <attrT> AttributeValue convertValue(
+            attrT attribute, TypeInformation<attrT> objectTypeInformation) {
+        if (attribute == null) {
+            return AttributeValue.builder().nul(true).build();
```
dannycranmer, line 106: Not sure this is the correct thing to do here? I believe a common practise is to omit null fields rather than setting to null. Maybe this could be a configuration on the element converter?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
```diff
@@ -0,0 +1,245 @@
...
+        }
+    }
+
+    private <attT> Map<String, AttributeValue> convertElementUsingTypeInfo(
+            attT t, CompositeType<attT> typeInfo) {
+        Map<String, AttributeValue> map = new HashMap<>();
+        for (String fieldKey : typeInfo.getFieldNames()) {
+            TypeInformation<attT> fieldType = typeInfo.getTypeAt(fieldKey);
+            try {
+                Field field = t.getClass().getDeclaredField(fieldKey);
```
dannycranmer, line 84: As per the [coding guidelines](https://flink.apache.org/how-to-contribute/code-style-and-quality-java/#java-reflection) we should not use reflection. Can we instead use the `TypeExtractor` class as suggested in the docs? 

Another note here is that loading the class information for each record is expensive. Suggest that we lazily cache the type info based on the class.

dannycranmer:
Please change commit message to include the component: `[FLINK-35022][Connectors/DynamoDB]`

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
```diff
@@ -0,0 +1,245 @@
...
+            Object attribute, TypeInformation<attrT> componentTypeInfo, boolean isPrimitive) {
+
+        if (componentTypeInfo.getTypeClass().isAssignableFrom(String.class)) {
+            return AttributeValue.fromSs(convertObjectArrayToStringList((Object[]) attribute));
+        } else if (componentTypeInfo instanceof NumericTypeInfo && !isPrimitive) {
+            return AttributeValue.fromNs(convertObjectArrayToStringList((Object[]) attribute));
+        } else if (componentTypeInfo instanceof NumericTypeInfo) {
+            if (attribute instanceof int[]) {
+                return AttributeValue.fromNs(
+                        (Arrays.stream((int[]) attribute)
```
dannycranmer, line 176: As a big fan of the stream API, it pains me to say that we [should not be using it for performance critical code](https://flink.apache.org/how-to-contribute/code-style-and-quality-java/#java-streams). Since this would be called for each record I would consider this performance critical/data-intensive. Please convert to standard loops, can extract to new methods for readability.

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverterTest.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.Map;
+
+import static org.apache.flink.connector.dynamodb.sink.DynamoDbWriteRequestType.PUT;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** Tests for {@link DynamoDbTypeInformedElementConverter}. */
+public class DynamoDbTypeInformedElementConverterTest {
+
+    @Test
+    public void testSimpleTypeConversion() {
```
dannycranmer, line 51: nit: JUnit 5 coding standards are slightly different from lower versions. In 5 we:
  1/ Use package protected class/methods (drop `public`)
  2/ Remove the `test` prefix on the methods (`testSimpleTypeConversion()` -> `simpleTypeConversion()`)

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverterTest.java
```diff
@@ -0,0 +1,161 @@
...
+        Throwable throwable =
+                Assertions.assertThrows(
+                        FlinkRuntimeException.class,
+                        () -> {
+                            elementConverter.apply(Tuple2.of(dateVal, "stringVal"), null);
+                        });
+
+        assertThat(throwable).hasCauseInstanceOf(IllegalArgumentException.class);
+        assertThat(throwable.getCause())
+                .hasMessageContaining("Unsupported BasicTypeInfo " + BasicTypeInfo.DATE_TYPE_INFO);
```
dannycranmer, line 140: Can you use the AssertJ lib here instead?

```
Assertions.assertThatExceptionOfType(FlinkRuntimeException.class)
                .isThrownBy(() -> elementConverter.apply(Tuple1.of("nan"), null))
                .withMessageContaining("Incompatible type when converting nan, using Integer type information.")
                .withRootCauseInstanceOf(IllegalArgumentException.class);
```

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
```diff
@@ -0,0 +1,245 @@
...
+            for (Object attributeItem : (Object[]) attribute) {
+                checkTypeCompatibility(attributeItem, componentTypeInfo);
+                attributeValueList.add(convertValue(((attrT) attributeItem), componentTypeInfo));
+            }
+            return AttributeValue.fromL(attributeValueList);
+        } else {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "Unsupported primitive typeInfo %s", componentTypeInfo.getTypeClass()));
+        }
```
dannycranmer, line 209: This block is untested, please add tests

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
```diff
@@ -0,0 +1,245 @@
...
+                    objectTypeInformation instanceof PrimitiveArrayTypeInfo);
+
+        } else if (objectTypeInformation instanceof TupleTypeInfo) {
+            TupleTypeInfo<?> tupleTypeInfo = (TupleTypeInfo<?>) objectTypeInformation;
+            Tuple t = (Tuple) attribute;
+            List<AttributeValue> attributeValues = new ArrayList<>();
+            for (int i = 0; i < tupleTypeInfo.getTotalFields(); i++) {
+                attributeValues.add(convertValue(t.getField(i), tupleTypeInfo.getTypeAt(i)));
+            }
+            return AttributeValue.fromL(attributeValues);
```
dannycranmer, line 147: This block is untested, please add tests

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
```diff
@@ -0,0 +1,245 @@
...
+                PrimitiveArrayTypeInfo.BYTE_PRIMITIVE_ARRAY_TYPE_INFO)) {
+            // special handling for byte array as it has special support in DynamoDB
+            if (attribute instanceof byte[]) {
+                return AttributeValue.fromB(SdkBytes.fromByteArray((byte[]) attribute));
+            } else if (attribute instanceof SdkBytes) {
+                return AttributeValue.fromB((SdkBytes) attribute);
+            } else {
+                throw new IllegalArgumentException(
+                        String.format("Failed to extract byte array type from %s", attribute));
+            }
```
dannycranmer, line 119: This block is untested, please add tests

dannycranmer (changes requested):

dannycranmer:
I have left a comment regarding the need for this on the Jira: https://issues.apache.org/jira/browse/FLINK-35022

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
vahmed-hamdy: @dannycranmer , yeah just checked `testConvertOrderToDynamoDbWriteRequestWithIgnoresNull`.

vahmed-hamdy:
@dannycranmer Thanks for the feedback, I have addressed your comments. Alternatively I decided to follow an approach similar to `RowDataToAttributeValueConverter` to avoid using reflection. PTAL

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
```diff
@@ -0,0 +1,380 @@
...
+        this.ignoreNulls = ignoreNulls;
+    }
+
+    @Override
+    public void open(Sink.InitContext context) {
+        try {
+            tableSchema = createTableSchema(typeInfo);
+        } catch (IntrospectionException | IllegalStateException | IllegalArgumentException e) {
+            throw new FlinkRuntimeException("Failed to extract DynamoDb table schema", e);
+        }
```
dannycranmer, line 95: We should do this in the constructor as a validation and fail fast. Do not need to store, but nice to fail on client and not when running job

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
```diff
@@ -0,0 +1,380 @@
...
+
+        return (TableSchema<attT>) tableSchemaBuilder.build();
+    }
+
+    private <attrT> TableSchema<attrT> createTableSchemaFromPojo(PojoTypeInfo<attrT> typeInfo)
+            throws IntrospectionException {
+        StaticTableSchema.Builder<attrT> tableSchemaBuilder =
+                StaticTableSchema.builder(typeInfo.getTypeClass());
+        BeanInfo beanInfo = Introspector.getBeanInfo(typeInfo.getTypeClass());
+        PropertyDescriptor[] propertyDescriptors = beanInfo.getPropertyDescriptors();
```
dannycranmer, line 169: This looks like it might be expensive, can we cache the descriptors per type?

dannycranmer (changes requested):
Thanks for the update @vahmed-hamdy. Left a few comments

dannycranmer:
@vahmed-hamdy can you also update the docs please?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
```diff
@@ -0,0 +1,380 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.function.Function;
+
+/**
+ * A {@link ElementConverter} that converts an element to a {@link DynamoDbWriteRequest} using
+ * TypeInformation provided.
+ */
+@PublicEvolving
+public class DynamoDbTypeInformedElementConverter<inputT>
```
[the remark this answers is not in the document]
vahmed-hamdy: I agree, I would suggest `DefaultDynamoDbElementConverter` since we are following up with making it default and changing it to optional? wdyt?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
vahmed-hamdy: I am not sure I get what we need to cache, these descriptors are only created once when creating the schema, during element conversion only the Enhanced getters are used.
Btw, this is also what the BeanTableSchema does :D

dannycranmer (approved):
LGTM, thanks @vahmed-hamdy

dannycranmer:
@hlteoh37 / @z3d1k can you please take a look at this one too?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
```diff
@@ -0,0 +1,380 @@
...
+            if (primitiveArrayTypeInfo.getComponentType() instanceof NumericTypeInfo) {
+                return getAttributeConverter(
+                        AttributeValueType.NS,
+                        array -> AttributeValue.fromNs(convertPrimitiveArrayToStringList(array)));
+            } else {
+                throw new IllegalArgumentException(
+                        String.format(
+                                "Unsupported primitive array typeInfo %s",
+                                primitiveArrayTypeInfo.getComponentType()));
+            }
```
hlteoh37, line 243: Have we considered using the `.attributeConverterProviders()` method in the table schema builder and reusing the constructs we use to DDB Table API?

```
TableSchema.builder(...)
                      .attributeConverterProviders(new ArrayAttributeConverterProvider(), ... (add any more here) ..., AttributeConverterProvider.defaultProvider());
```

We have an example here:
https://github.com/apache/flink-connector-aws/blob/main/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverter.java#L56-L72

hlteoh37 (changes requested):
Thanks for the PR @vahmed-hamdy ! Added a comment

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
vahmed-hamdy: @hlteoh37 yes, the whole approach actually is very similar, however The way `AttributeConverterProvider` works is by mapping `EnhancedType` provided to an `AttributeConverter`, We here rely on Flink's `TypeInfo` rather than the `EnhancedType` that's why we created an alternative structure. Wdyt?

vahmed-hamdy:
@hlteoh37 Thanks, for the feedback, the reason we took this approach is that We are trying to couple it as much as possible with Flink's TypeInfo Class, Using `AttributeConverterProvider` is closer to DDB's `EnhancedType` rather than Flink's

vahmed-hamdy:
@hlteoh37 Do you have other comments or could we merge this PR?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
hlteoh37: Hm, ultimately `EnhancedType` is just a wrapper for the `AttributeConverterProvider` to match the right `AttributeConverter`. We cannot run away from this - in fact you have created a custom implementation of this logic in `getAttributeConverter`, which takes type information and constructs a custom attribute converter. 

Wouldn't using the `AttributeConverterProvider` more directly simplify the code?

hlteoh37:
Hi @vahmed-hamdy, my preference would be for us to refactor the code in `DynamoDbTypeInformedElementConverter` to use a cleaner structure rather than a long list of `if-else`. Is this something we need to close urgently?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/sink/DynamoDbTypeInformedElementConverter.java
vahmed-hamdy: > We cannot run away from this 

This is not true, we are "running away from this" or rather replacing the mapping dependency on `EnhancedType` by directly attaching each resolved `AttribureConverter` to the attribute in
```
  typeInfo.getTypeClass(),
                a ->
                        a.name(fieldName)
                                .getter(getter)
                                .setter((e, o) -> {})
                                .attributeConverter(getAttributeConverter(typeInfo)))
```

> in fact you have created a custom implementation of this logic in getAttributeConverter, which takes type information and constructs a custom attribute converter.

Yes, I am aware of that and that exactly why using an attribute converter is against what we are trying to achieve here, Are you suggesting we don't depend on Flink's TypeInformation and resolve back to Java Classes?

hlteoh37 (approved):
Thanks @vahmed-hamdy

PR #139, opened 2024-04-29
[FLINK-35269] Log information about failed records as warnings instead of debug for Kinesis and Firehose sinks

z3d1k:
## Purpose of the change

Change logging level for information about records failed to be written by Kinesis and Firehose sinks from `debug` to `warn`.
Similar configuration is already in use for DynamoDB sink.

## Verifying this change

## Significant changes

### DIFF
--- a/flink-connector-aws/flink-connector-aws-kinesis-firehose/src/main/java/org/apache/flink/connector/firehose/sink/KinesisFirehoseSinkWriter.java
+++ b/flink-connector-aws/flink-connector-aws-kinesis-firehose/src/main/java/org/apache/flink/connector/firehose/sink/KinesisFirehoseSinkWriter.java
@@ -220,9 +220,8 @@ class KinesisFirehoseSinkWriter<InputT> extends AsyncSinkWriter<InputT, Record> 
         }
 
         LOG.warn(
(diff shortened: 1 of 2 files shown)


flink-connector-aws/flink-connector-aws-kinesis-firehose/src/main/java/org/apache/flink/connector/firehose/sink/KinesisFirehoseSinkWriter.java
```diff
@@ -220,9 +220,8 @@ private void handleFullyFailedRequest(
         }
 
         LOG.warn(
-                "KDF Sink failed to write and will retry {} entries to KDF first request was {}",
+                "KDF Sink failed to write and will retry {} entries to KDF",
```
hlteoh37, line 223: Can we explain here why we remove the logging for first entry in request?

z3d1k: Logging the first entry in warn log here has several issues:
- records can be very big, bloating logs
- records are in serialised format (i.e. byte data), which has questionable value.
If this data is required, this can be logged separately at debug level to control log granularity

PR #151, opened 2024-07-18
[FLINK-35660][Connectors/DynamoDB] Adding parent-child shard ordering for DDB Streams source

gguptp:
## Purpose of the change

- Implements parent-child shard ordering in DDB Streams source
- Source enumerator will only assign splits that correspond to a shard when parent shard is either finished, expired or doesn't exist
- Source enumerator also discovers inconsistencies in the shard graph due to describestream and fixes it
- Implements millisBehindLatest metric for DDB Streams source

## Verifying this change

Added unit tests

## Significant changes

### DIFF
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/config/DynamodbStreamsSourceConfigConstants.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/config/DynamodbStreamsSourceConfigConstants.java
@@ -22,6 +22,8 @@ import org.apache.flink.annotation.Experimental;
 import org.apache.flink.configuration.ConfigOption;
 import org.apache.flink.configuration.ConfigOptions;
 
+import java.time.Duration;
+
 /** Constants to be used with the DynamodbStreamsSource. */
 @Experimental
 public class DynamodbStreamsSourceConfigConstants {
@@ -37,11 +39,39 @@ public class DynamodbStreamsSourceConfigConstants {
                     .defaultValue(InitialPosition.LATEST)
                     .withDescription("The initial position to start reading Dynamodb streams.");
 
-    public static final ConfigOption<Long> SHARD_DISCOVERY_INTERVAL_MILLIS =
+    public static final ConfigOption<Duration> SHARD_DISCOVERY_INTERVAL =
             ConfigOptions.key("flink.shard.discovery.intervalmillis")
-                    .longType()
-                    .defaultValue(10000L)
+                    .durationType()
+                    .defaultValue(Duration.ofSeconds(60))
                     .withDescription("The interval between each attempt to discover new shards.");
+
+    public static final ConfigOption<Integer> DESCRIBE_STREAM_INCONSISTENCY_RESOLUTION_RETRY_COUNT =
+            ConfigOptions.key("flink.describestream.inconsistencyresolution.retries")
+                    .intType()
+                    .defaultValue(5)
+                    .withDescription(
+                            "The number of times to retry build shard lineage if describestream returns inconsistent response");
+
+    public static final ConfigOption<Integer> DYNAMODB_STREAMS_RETRY_COUNT =
+            ConfigOptions.key("flink.dynamodbstreams.numretries")
+                    .intType()
+                    .defaultValue(50)
+                    .withDescription(
+                            "The number of times to retry DynamoDB Streams API call if it returns a retryable exception");
+
+    public static final ConfigOption<Duration> DYNAMODB_STREAMS_EXPONENTIAL_BACKOFF_MIN_DELAY =
+            ConfigOptions.key("flink.dynamodbstreams.backoff.mindelayduration")
+                    .durationType()
+                    .defaultValue(Duration.ofMillis(100))
+                    .withDescription(
+                            "The minimum delay for exponential backoff for describestream");
+
+    public static final ConfigOption<Duration> DYNAMODB_STREAMS_EXPONENTIAL_BACKOFF_MAX_DELAY =
+            ConfigOptions.key("flink.dynamodbstreams.backoff.maxdelay")
+                    .durationType()
+                    .defaultValue(Duration.ofMillis(1000))
+                    .withDescription(
+                            "The maximum delay for exponential backoff for describestream");
 
     public static final String BASE_DDB_STREAMS_USER_AGENT_PREFIX_FORMAT =
             "Apache Flink %s (%s) DynamoDb Streams Connector";
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
@@ -20,19 +20,24 @@ package org.apache.flink.connector.dynamodb.source.enumerator;
 
 import org.apache.flink.annotation.Internal;
 import org.apache.flink.api.connector.source.ReaderInfo;
+import org.apache.flink.api.connector.source.SourceEvent;
 import org.apache.flink.api.connector.source.SplitEnumerator;
 import org.apache.flink.api.connector.source.SplitEnumeratorContext;
 import org.apache.flink.api.connector.source.SplitsAssignment;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.connector.dynamodb.source.config.DynamodbStreamsSourceConfigConstants.InitialPosition;
+import org.apache.flink.connector.dynamodb.source.enumerator.event.SplitsFinishedEvent;
+import org.apache.flink.connector.dynamodb.source.enumerator.tracker.SplitGraphInconsistencyTracker;
+import org.apache.flink.connector.dynamodb.source.enumerator.tracker.SplitTracker;
 import org.apache.flink.connector.dynamodb.source.exception.DynamoDbStreamsSourceException;
 import org.apache.flink.connector.dynamodb.source.proxy.StreamProxy;
 import org.apache.flink.connector.dynamodb.source.split.DynamoDbStreamsShardSplit;
-import org.apache.flink.connector.dynamodb.source.split.StartingPosition;
+import org.apache.flink.connector.dynamodb.source.util.ListShardsResult;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.services.dynamodb.model.Shard;
+import software.amazon.awssdk.services.dynamodb.model.StreamStatus;
 
 import javax.annotation.Nullable;
 
@@ -45,8 +50,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.stream.Collectors;
 
-import static org.apache.flink.connector.dynamodb.source.config.DynamodbStreamsSourceConfigConstants.SHARD_DISCOVERY_INTERVAL_MILLIS;
+import static org.apache.flink.connector.dynamodb.source.config.DynamodbStreamsSourceConfigConstants.DESCRIBE_STREAM_INCONSISTENCY_RESOLUTION_RETRY_COUNT;
+import static org.apache.flink.connector.dynamodb.source.config.DynamodbStreamsSourceConfigConstants.SHARD_DISCOVERY_INTERVAL;
 import static org.apache.flink.connector.dynamodb.source.config.DynamodbStreamsSourceConfigConstants.STREAM_INITIAL_POSITION;
 
 /**
@@ -67,12 +74,9 @@ public class DynamoDbStreamsSourceEnumerator
     private final StreamProxy streamProxy;
     private final DynamoDbStreamsShardAssigner shardAssigner;
     private final ShardAssignerContext shardAssignerContext;
+    private final SplitTracker splitTracker;
 
     private final Map<Integer, Set<DynamoDbStreamsShardSplit>> splitAssignment = new HashMap<>();
-    private final Set<String> assignedSplitIds = new HashSet<>();
-    private final Set<DynamoDbStreamsShardSplit> unassignedSplits;
-
-    private String lastSeenShardId;
 
     public DynamoDbStreamsSourceEnumerator(
             SplitEnumeratorContext<DynamoDbStreamsShardSplit> context,
@@ -87,25 +91,22 @@ public class DynamoDbStreamsSourceEnumerator
         this.streamProxy = streamProxy;
         this.shardAssigner = shardAssigner;
         this.shardAssignerContext = new ShardAssignerContext(splitAssignment, context);
+        InitialPosition initialPosition = sourceConfig.get(STREAM_INITIAL_POSITION);
         if (state == null) {
-            this.lastSeenShardId = null;
-            this.unassignedSplits = new HashSet<>();
+            this.splitTracker = new SplitTracker(streamArn, initialPosition);
         } else {
-            this.lastSeenShardId = state.getLastSeenShardId();
-            this.unassignedSplits = state.getUnassignedSplits();
+            this.splitTracker =
+                    new SplitTracker(state.getKnownSplits(), streamArn, initialPosition);
         }
     }
 
     @Override
     public void start() {
-        if (lastSeenShardId == null) {
-            context.callAsync(this::initialDiscoverSplits, this::assignSplits);
-        }
-
-        final long shardDiscoveryInterval = sourceConfig.get(SHARD_DISCOVERY_INTERVAL_MILLIS);
+        context.callAsync(this::discoverSplits, this::processDiscoveredSplits);
+        final long shardDiscoveryInterval = sourceConfig.get(SHARD_DISCOVERY_INTERVAL).toMillis();
         context.callAsync(
-                this::periodicallyDiscoverSplits,
-                this::assignSplits,
+                this::discoverSplits,
+                this::processDiscoveredSplits,
                 shardDiscoveryInterval,
                 shardDiscoveryInterval);
     }
@@ -116,24 +117,96 @@ public class DynamoDbStreamsSourceEnumerator
     }
 
     @Override
-    public void addSplitsBack(List<DynamoDbStreamsShardSplit> splits, int subtaskId) {
-        if (!splitAssignment.containsKey(subtaskId)) {
-            LOG.warn(
-                    "Unable to add splits back for subtask {} since it is not assigned any splits. Splits: {}",
-                    subtaskId,
-                    splits);
+    public void addSplitsBack(List<DynamoDbStreamsShardSplit> list, int i) {
+        throw new UnsupportedOperationException("Partial recovery is not supported");
+    }
+
+    @Override
+    public void handleSourceEvent(int subtaskId, SourceEvent sourceEvent) {
+        if (sourceEvent instanceof SplitsFinishedEvent) {
+            handleFinishedSplits(subtaskId, (SplitsFinishedEvent) sourceEvent);
+        }
+    }
+
+    private void handleFinishedSplits(int subtaskId, SplitsFinishedEvent splitsFinishedEvent) {
+        splitTracker.markAsFinished(splitsFinishedEvent.getFinishedSplitIds());
+        splitAssignment
+                .get(subtaskId)
+                .removeIf(
+                        split ->
+                                splitsFinishedEvent
+                                        .getFinishedSplitIds()
+                                        .contains(split.splitId()));
+        assignSplits();
+    }
+
+    private void processDiscoveredSplits(ListShardsResult discoveredSplits, Throwable throwable) {
+        if (throwable != null) {
+            throw new DynamoDbStreamsSourceException("Failed to list shards.", throwable);
+        }
+
+        if (discoveredSplits.getInconsistencyDetected()) {
             return;
         }
 
-        for (DynamoDbStreamsShardSplit split : splits) {
-            splitAssignment.get(subtaskId).remove(split);
-            assignedSplitIds.remove(split.splitId());
-            unassignedSplits.add(split);
+        splitTracker.addSplits(discoveredSplits.getShards());
+        splitTracker.cleanUpOldFinishedSplits(
+                discoveredSplits.getShards().stream()
+                        .map(Shard::shardId)
+                        .collect(Collectors.toSet()));
+        if (context.registeredReaders().size() < context.currentParallelism()) {
+            LOG.info(
+                    "Insufficient registered readers, skipping assignment of discovered splits until all readers are registered. Required number of readers: {}, registered readers: {}",
+                    context.currentParallelism(),
+                    context.registeredReaders().size());
+            return;
         }
+        assignSplits();
+    }
 
-        // Assign the unassignedSplits
-        // We did not discover any new splits, so we put in an empty list
-        assignSplits(Collections.emptyList(), null);
+    /**
+     * This method tracks the discovered splits in a graph and if the graph has inconsistencies, it
+     * tries to resolve them using DescribeStream calls using the first inconsistent node found in
+     * the split graph.
+     *
+     * @param discoveredSplits splits discovered after calling DescribeStream at the start of the
+     *     application or periodically.
+     */
+    private SplitGraphInconsistencyTracker trackSplitsAndResolveInconsistencies(
+            ListShardsResult discoveredSplits) {
+        SplitGraphInconsistencyTracker splitGraphInconsistencyTracker =
+                new SplitGraphInconsistencyTracker();
+        splitGraphInconsistencyTracker.addNodes(discoveredSplits.getShards());
+
+        // we don't want to do inconsistency checks for DISABLED streams because there will be no
+        // open child shard in DISABLED stream
+        boolean streamDisabled = discoveredSplits.getStreamStatus().equals(StreamStatus.DISABLED);
+        int describeStreamInconsistencyResolutionCount =
+                sourceConfig.get(DESCRIBE_STREAM_INCONSISTENCY_RESOLUTION_RETRY_COUNT);
+        for (int i = 0;
+                i < describeStreamInconsistencyResolutionCount
+                        && !streamDisabled
+                        && splitGraphInconsistencyTracker.inconsistencyDetected();
+                i++) {
+            String earliestClosedLeafNodeId =
+                    splitGraphInconsistencyTracker.getEarliestClosedLeafNode();
+            LOG.warn(
+                    "We have detected inconsistency with DescribeStream output, resolving inconsistency with shardId: {}",
+                    earliestClosedLeafNodeId);
+            ListShardsResult shardsToResolveInconsistencies =
+                    streamProxy.listShards(streamArn, earliestClosedLeafNodeId);
+            splitGraphInconsistencyTracker.addNodes(shardsToResolveInconsistencies.getShards());
+        }
+        return splitGraphInconsistencyTracker;
+    }
+
+    private void assignSplits() {
+        Map<Integer, List<DynamoDbStreamsShardSplit>> newSplitAssignments = new HashMap<>();
+        for (DynamoDbStreamsShardSplit split : splitTracker.splitsAvailableForAssignment()) {
+            assignSplitToSubtask(split, newSplitAssignments);
+        }
+        updateSplitAssignment(newSplitAssignments);
+        context.assignSplits(new SplitsAssignment<>(newSplitAssignments));
     }
 
     @Override
@@ -143,17 +216,14 @@ public class DynamoDbStreamsSourceEnumerator
 
     @Override
     public DynamoDbStreamsSourceEnumeratorState snapshotState(long checkpointId) throws Exception {
-        return new DynamoDbStreamsSourceEnumeratorState(unassignedSplits, lastSeenShardId);
+        List<DynamoDBStreamsShardSplitWithAssignmentStatus> splitStates =
+                splitTracker.snapshotState(checkpointId);
+        return new DynamoDbStreamsSourceEnumeratorState(splitStates);
     }
 
     @Override
     public void close() throws IOException {
         streamProxy.close();
-    }
-
-    private List<DynamoDbStreamsShardSplit> initialDiscoverSplits() {
-        List<Shard> shards = streamProxy.listShards(streamArn, lastSeenShardId);
-        return mapToSplits(shards, sourceConfig.get(STREAM_INITIAL_POSITION));
     }
 
     /**
@@ -162,76 +232,35 @@ public class DynamoDbStreamsSourceEnumerator
      *
      * @return list of discovered splits
      */
-    private List<DynamoDbStreamsShardSplit> periodicallyDiscoverSplits() {
-        List<Shard> shards = streamProxy.listShards(streamArn, lastSeenShardId);
+    private ListShardsResult discoverSplits() {
+        ListShardsResult listShardsResult = streamProxy.listShards(streamArn, null);
+        SplitGraphInconsistencyTracker splitGraphInconsistencyTracker =
+                trackSplitsAndResolveInconsistencies(listShardsResult);
 
-        // Any shard discovered after the initial startup should be read from the start, since they
-        // come from resharding
-        return mapToSplits(shards, InitialPosition.TRIM_HORIZON);
-    }
-
-    private List<DynamoDbStreamsShardSplit> mapToSplits(
-            List<Shard> shards, InitialPosition initialPosition) {
-        StartingPosition startingPosition;
-        switch (initialPosition) {
-            case LATEST:
-                startingPosition = StartingPosition.latest();
-                break;
-            case TRIM_HORIZON:
-            default:
-                startingPosition = StartingPosition.fromStart();
+        ListShardsResult discoveredSplits = new ListShardsResult();
+        discoveredSplits.setStreamStatus(listShardsResult.getStreamStatus());
+        discoveredSplits.setInconsistencyDetected(listShardsResult.getInconsistencyDetected());
+        List<Shard> shardList = new ArrayList<>(splitGraphInconsistencyTracker.getNodes());
+        // We do not throw an exception here and just return to let SplitTracker process through the
+        // splits it has not yet processed. This might be helpful for large streams which see a lot
+        // of
+        // inconsistency issues.
+        if (splitGraphInconsistencyTracker.inconsistencyDetected()) {
+            LOG.error(
+                    "There are inconsistencies in DescribeStream which we were not able to resolve. First leaf node on which inconsistency was detected:"
+                            + splitGraphInconsistencyTracker.getEarliestClosedLeafNode());
+            return discoveredSplits;
         }
-
-        List<DynamoDbStreamsShardSplit> splits = new ArrayList<>();
-        for (Shard shard : shards) {
-            splits.add(new DynamoDbStreamsShardSplit(streamArn, shard.shardId(), startingPosition));
-        }
-
-        return splits;
-    }
-
-    /**
-     * This method assigns a given set of DynamoDb Streams splits to the readers currently
-     * registered on the cluster. This assignment is done via a side-effect on the {@link
-     * SplitEnumeratorContext} object.
-     *
-     * @param discoveredSplits list of discovered splits
-     * @param throwable thrown when discovering splits. Will be null if no throwable thrown.
-     */
-    private void assignSplits(
-            List<DynamoDbStreamsShardSplit> discoveredSplits, Throwable throwable) {
-        if (throwable != null) {
-            throw new DynamoDbStreamsSourceException("Failed to list shards.", throwable);
-        }
-
-        if (context.registeredReaders().size() < context.currentParallelism()) {
-            LOG.info(
-                    "Insufficient registered readers, skipping assignment of discovered splits until all readers are registered. Required number of readers: {}, Registered readers: {}",
-                    context.currentParallelism(),
-                    context.registeredReaders().size());
-            unassignedSplits.addAll(discoveredSplits);
-            return;
-        }
-
-        Map<Integer, List<DynamoDbStreamsShardSplit>> newSplitAssignments = new HashMap<>();
-        for (DynamoDbStreamsShardSplit split : unassignedSplits) {
-            assignSplitToSubtask(split, newSplitAssignments);
-        }
-        unassignedSplits.clear();
-        for (DynamoDbStreamsShardSplit split : discoveredSplits) {
-            assignSplitToSubtask(split, newSplitAssignments);
-        }
-
-        updateLastSeenShardId(discoveredSplits);
-        updateSplitAssignment(newSplitAssignments);
-        context.assignSplits(new SplitsAssignment<>(newSplitAssignments));
+        discoveredSplits.addShards(shardList);
+        return discoveredSplits;
     }
 
     private void assignSplitToSubtask(
             DynamoDbStreamsShardSplit split,
             Map<Integer, List<DynamoDbStreamsShardSplit>> newSplitAssignments) {
-        if (assignedSplitIds.contains(split.splitId())) {
-            LOG.info(
+        if (splitTracker.isAssigned(split.splitId())) {
+
+            LOG.warn(
                     "Skipping assignment of shard {} from stream {} because it is already assigned.",
                     split.getShardId(),
                     split.getStreamArn());
@@ -255,14 +284,7 @@ public class DynamoDbStreamsSourceEnumerator
             subtaskList.add(split);
             newSplitAssignments.put(selectedSubtask, subtaskList);
         }
-        assignedSplitIds.add(split.splitId());
-    }
-
-    private void updateLastSeenShardId(List<DynamoDbStreamsShardSplit> discoveredSplits) {
-        if (!discoveredSplits.isEmpty()) {
-            DynamoDbStreamsShardSplit lastSplit = discoveredSplits.get(discoveredSplits.size() - 1);
-            lastSeenShardId = lastSplit.getShardId();
-        }
+        splitTracker.markAsAssigned(Collections.singletonList(split));
     }
 
     private void updateSplitAssignment(
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumeratorStateSerializer.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumeratorStateSerializer.java
@@ -29,14 +29,13 @@ import java.io.ByteArrayOutputStream;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
-import java.util.HashSet;
-import java.util.Set;
+import java.util.ArrayList;
+import java.util.List;
 
 /** Used to serialize and deserialize the {@link DynamoDbStreamsSourceEnumeratorState}. */
 @Internal
 public class DynamoDbStreamsSourceEnumeratorStateSerializer
         implements SimpleVersionedSerializer<DynamoDbStreamsSourceEnumeratorState> {
-
     private static final int CURRENT_VERSION = 0;
 
     private final DynamoDbStreamsShardSplitSerializer splitSerializer;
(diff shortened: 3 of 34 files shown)


flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
```diff
@@ -0,0 +1,329 @@
...
+            String streamArn,
+            DynamodbStreamsSourceConfigConstants.InitialPosition initialPosition) {
+        this.streamArn = streamArn;
+        this.initialPosition = initialPosition;
+        initialState.forEach(
+                splitWithStatus -> {
+                    DynamoDbStreamsShardSplit currentSplit = splitWithStatus.split();
+                    knownSplits.put(currentSplit.splitId(), currentSplit);
+
+                    if (ASSIGNED.equals(splitWithStatus.assignmentStatus())) {
```
karubian, line 73: nit: Is it possible to get a null value from `assignmentStatus`? We can consider converting the Yoda convention in that case.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/config/DynamodbStreamsSourceConfigConstants.java
```diff
@@ -43,6 +45,15 @@ public enum InitialPosition {
...
                     .withDescription("The interval between each attempt to discover new shards.");
 
+    public static final ConfigOption<Integer> DESCRIBE_STREAM_INCONSISTENCY_RESOLUTION_RETRY_COUNT =
+            ConfigOptions.key("flink.describestream.inconsistencyresolution.retries")
+                    .intType()
+                    .defaultValue(5)
+                    .withDescription(
+                            "The number of times to retry build shard lineage if describestream returns inconsistent response");
+
+    public static final Duration MIN_DDB_STREAMS_SHARD_RETENTION = Duration.ofHours(6);
```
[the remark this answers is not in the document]
gguptp: Nope, this shouldn't be defined via config

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
```diff
@@ -116,24 +117,76 @@ public void handleSplitRequest(int subtaskId, @Nullable String requesterHostname
...
+
+        ShardGraphTracker shardGraphTracker = new ShardGraphTracker();
+        shardGraphTracker.addNodes(discoveredSplits);
+
+        StreamDescription streamDescription = streamProxy.getStreamDescription(streamArn, null);
+        boolean streamDisabled = streamDescription.streamStatus().equals(StreamStatus.DISABLED);
+
+        for (int i = 0;
+                i < sourceConfig.getInteger(DESCRIBE_STREAM_INCONSISTENCY_RESOLUTION_RETRY_COUNT)
+                        && !streamDisabled
```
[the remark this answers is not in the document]
gguptp: good point, will change this

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
gguptp: will check this, although i believe it should not be null

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
```diff
@@ -162,75 +216,14 @@ private List<DynamoDbStreamsShardSplit> initialDiscoverSplits() {
...
+        return streamProxy.listShards(streamArn, null);
     }
 
     private void assignSplitToSubtask(
             DynamoDbStreamsShardSplit split,
             Map<Integer, List<DynamoDbStreamsShardSplit>> newSplitAssignments) {
-        if (assignedSplitIds.contains(split.splitId())) {
+        if (splitTracker.isAssigned(split.splitId())) {
             LOG.info(
                     "Skipping assignment of shard {} from stream {} because it is already assigned.",
```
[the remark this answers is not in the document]
gguptp: Made it to debug

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
gguptp: We wont get null here, since during state initialization, we are making sure that this value is filled always and when we are putting it to state, we are making sure that this has some value

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
```diff
@@ -116,24 +117,80 @@ public void handleSplitRequest(int subtaskId, @Nullable String requesterHostname
...
+        if (throwable != null) {
+            throw new DynamoDbStreamsSourceException("Failed to list shards.", throwable);
+        }
+
+        ShardGraphTracker shardGraphTracker = new ShardGraphTracker();
+        shardGraphTracker.addNodes(discoveredSplits.getShards());
+
+        boolean streamDisabled = discoveredSplits.getStreamStatus().equals(StreamStatus.DISABLED);
+        int describeStreamInconsistencyResolutionCount =
+                sourceConfig.getInteger(DESCRIBE_STREAM_INCONSISTENCY_RESOLUTION_RETRY_COUNT);
```
hlteoh37, line 153: can we abstract this away to at least a `private` method in this class, or even better a separate class like `DDBStreamConsistencyTracker`? Would make the code for `processDiscoveredSplits` easier to read (and test, if it's in a separate class)

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
```diff
@@ -116,24 +117,80 @@ public void handleSplitRequest(int subtaskId, @Nullable String requesterHostname
...
+                    streamProxy.listShards(
+                            streamArn, shardGraphTracker.getEarliestClosedLeafNode());
+            shardGraphTracker.addNodes(shardsToResolveInconsistencies.getShards());
+        }
+
+        if (shardGraphTracker.inconsistencyDetected()) {
+            LOG.error(
+                    "There are inconsistencies in describestream which we were not able to resolve. First leaf node on which inconsistency was detected:"
+                            + shardGraphTracker.getEarliestClosedLeafNode());
             return;
```
hlteoh37, line 169: Would be good to add a reason why we return here and not throw and exception.

I think we should also mention it on the description of `DESCRIBE_STREAM_INCONSISTENCY_RESOLUTION_RETRY_COUNT` and also the connector overview!

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
```diff
@@ -116,24 +117,80 @@ public void handleSplitRequest(int subtaskId, @Nullable String requesterHostname
...
         }
 
-        for (DynamoDbStreamsShardSplit split : splits) {
-            splitAssignment.get(subtaskId).remove(split);
-            assignedSplitIds.remove(split.splitId());
-            unassignedSplits.add(split);
+        splitTracker.addSplits(shardGraphTracker.getNodes());
```
hlteoh37, line 172: This is a little confusing - can we name the `splitGraphTracker` something like `splitGraphManager`?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
```diff
@@ -116,24 +117,80 @@ public void handleSplitRequest(int subtaskId, @Nullable String requesterHostname
...
+        splitTracker.cleanupGarbageSplits(
+                shardGraphTracker.getNodes().stream()
+                        .map(shard -> shard.shardId())
+                        .collect(Collectors.toSet()));
```
hlteoh37, line 176: nit: should the method just be `removeSplits()`?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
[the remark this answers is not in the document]
hlteoh37: Do we expect this to be printed constantly, or just when there are data inconsistencies? If this is only printed during data inconsistencies, then we should keep it as `INFO`

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumeratorStateSerializer.java
```diff
@@ -29,15 +29,19 @@
...
+import java.util.List;
 import java.util.Set;
 
 /** Used to serialize and deserialize the {@link DynamoDbStreamsSourceEnumeratorState}. */
 @Internal
 public class DynamoDbStreamsSourceEnumeratorStateSerializer
         implements SimpleVersionedSerializer<DynamoDbStreamsSourceEnumeratorState> {
 
-    private static final int CURRENT_VERSION = 0;
+    private static final Set<Integer> COMPATIBLE_VERSIONS = new HashSet<>(Arrays.asList(0, 1));
```
hlteoh37, line 43: we have not actually released this connector - so we don't need to up the serializer version even though we are making backwards incompatible changes. This is different from the KDS connector. Can we keep the `CURRENT_VERSION` as 0?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/ShardGraphTracker.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeSet;
+
+/**
+ * Class to track the state of shard graph created as part of describestream operation. This will
+ * track for inconsistent shards returned due to describestream operation. Caller will have to call
+ * describestream again with the chronologically first leaf node to resolve inconsistencies.
```
hlteoh37, line 32: Q: Should we make this logic part of this class? e.g. rename as `ShardGraphConsistencyDetector`? Not 100% sure of this suggestion. WDYT @gguptp @karubian

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/split/DynamoDbStreamsShardSplitSerializer.java
```diff
@@ -93,8 +100,19 @@ public DynamoDbStreamsShardSplit deserialize(int version, byte[] serialized)
                 }
             }
 
+            Set<String> parentShardIds = new HashSet<>();
+            if (version == CURRENT_VERSION) {
```
hlteoh37, line 104: This check is already done on line 83

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/ShardGraphTrackerTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Test
+    public void testShardGraphTrackerDetectsInconsistencies() {
+        List<Shard> shards =
+                Arrays.asList(
+                        // shards which don't have a parent
+                        generateShard(0, "1400", "1700", null),
+                        generateShard(1, "1500", "1800", null),
+                        // shards produced by rotation of parents
+                        generateShard(2, "1710", null, generateShardId(0)));
```
hlteoh37, line 60: Can we put in a later `closedLeafNode` so that we can test that the earlier one is detected? Best in a new test

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/ShardGraphTrackerTest.java
```diff
@@ -0,0 +1,67 @@
...
+                        // shards which don't have a parent
+                        generateShard(0, "1400", "1700", null),
+                        generateShard(1, "1500", "1800", null),
+                        // shards produced by rotation of parents
+                        generateShard(2, "1710", null, generateShardId(0)),
+                        generateShard(3, "1520", null, generateShardId(1)));
+        SplitGraphInconsistencyTracker shardGraphTracker = new SplitGraphInconsistencyTracker();
+        shardGraphTracker.addNodes(shards);
+        assertThat(shardGraphTracker.inconsistencyDetected()).isFalse();
+        assertThat(shardGraphTracker.getNodes()).containsExactlyInAnyOrderElementsOf(shards);
```
hlteoh37, line 49: Can we validate the behavior of other public APIs like `getEarliestClosedLeafNode()`?

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/source/util/DynamoDbStreamsProxyProvider.java
```diff
@@ -56,17 +58,19 @@ public static class TestDynamoDbStreamsProxy implements StreamProxy {
 
         // List shards configuration
         private final List<Shard> shards = new ArrayList<>();
+        private final Instant creationTimestamp = Instant.now();
         private Supplier<Exception> listShardsExceptionSupplier;
-        private boolean shouldRespectLastSeenShardId = true;
+        private boolean shouldRespectLastSeenShardId = false;
```
hlteoh37, line 63: we can just remove support for this

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/proxy/DynamoDbStreamsProxy.java
```diff
@@ -62,7 +61,7 @@ public DynamoDbStreamsProxy(
     }
 
     @Override
-    public List<Shard> listShards(String streamArn, @Nullable String lastSeenShardId) {
+    public ListShardsResult listShards(String streamArn, @Nullable String lastSeenShardId) {
```
hlteoh37, line 64: can we remove the lastSeenShardId field here?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
```diff
@@ -143,17 +217,18 @@ public void addReader(int subtaskId) {
 
     @Override
     public DynamoDbStreamsSourceEnumeratorState snapshotState(long checkpointId) throws Exception {
```
hlteoh37, line 219: This method is not covered in unit tests

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
```diff
@@ -116,24 +118,96 @@ public void handleSplitRequest(int subtaskId, @Nullable String requesterHostname
...
-                    splits);
+    public void addSplitsBack(List<DynamoDbStreamsShardSplit> list, int i) {
+        throw new UnsupportedOperationException("Partial recovery is not supported");
+    }
+
+    @Override
+    public void handleSourceEvent(int subtaskId, SourceEvent sourceEvent) {
+        if (sourceEvent instanceof SplitsFinishedEvent) {
+            handleFinishedSplits(subtaskId, (SplitsFinishedEvent) sourceEvent);
+        }
```
hlteoh37, line 129: This is not unit tested!

hlteoh37 (changes requested):
Thanks for the contribution @gguptp !

Added some comments - and have discussed offline.

Also measured test coverage as below

image

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
```diff
@@ -0,0 +1,252 @@
...
+        if (parentShard == null) {
+            return TRIM_HORIZON;
+        }
+        InitialPosition parentStartingPosition =
+                getStartingPosition(parentShard, shardIdToShardMap);
+        if (parentStartingPosition == initialPosition) {
+            return TRIM_HORIZON;
+        }
+        return parentStartingPosition;
+    }
```
hlteoh37, line 112: I don't understand why we need to resolve this starting position here. Since we only support `LATEST` and `TRIM_HORIZON`, could we not do the below logic:

1. If `TRIM_HORIZON` configured, every shard will be `TRIM_HORIZON`.
2. If `LATEST` configured, we filter such that we only consume active shards, and every shard will be `LATEST`

hlteoh37 (changes requested):
Thanks for the iteration @gguptp! The `SplitTracker` looks much cleaner!. There were some comments that haven't been addressed, and added some comments on the `SplitTracker`.

gguptp:
Thanks for the review @hlteoh37! I updated the PR after addressing the comments.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
gguptp: Yeah, makes sense. Updated

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
```diff
@@ -116,24 +118,96 @@ public void handleSplitRequest(int subtaskId, @Nullable String requesterHostname
...
+    /**
+     * This method tracks the discovered splits in a graph and if the graph has inconsistencies, it
+     * tries to resolve them using DescribeStream calls using the first inconsistent node found in
+     * the split graph.
+     *
+     * @param discoveredSplits splits discovered after calling DescribeStream at the start of the
+     *     application or periodically.
+     */
+    private SplitGraphInconsistencyTracker trackSplitsAndResolveInconsistencies(
+            ListShardsResult discoveredSplits) {
```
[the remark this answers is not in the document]
hlteoh37: Discussed offline - would be good for us to move it to `discoverSplits()` as otherwise it will block

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
```diff
@@ -0,0 +1,237 @@
...
+            default:
+                startingPosition = StartingPosition.fromStart();
+        }
+
+        Set<String> parentShardIds = new HashSet<>();
+        if (shard.parentShardId() != null) {
+            parentShardIds.add(shard.parentShardId());
+        }
+        return new DynamoDbStreamsShardSplit(
+                streamArn, shard.shardId(), startingPosition, parentShardIds);
```
hlteoh37, line 124: If we know that DDB shards only have 1 parent, maybe let's remove the `Set` and instead just pass the parentShardId directly into the `DynamoDbStreamsShardSplit`?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    private InitialPosition getStartingPosition(Shard shard) {
+        if (shard.parentShardId() == null) {
+            return initialPosition;
+        }
+        if (!knownSplits.containsKey(shard.parentShardId())) {
+            return initialPosition;
+        }
+        return TRIM_HORIZON;
```
hlteoh37, line 104: Sorry to harp on this again, but I don't really get this logic. Why do we return `initialPosition` (which can be `TRIM_HORIZON` or `LATEST` if there is no `parentShard`? Wouldn't this mean that if a customer starts from `LATEST` in that case, wouldn't it also read from the `TRIM_HORIZON` shard?

Wonder if we can simplify the logic for `shardDiscovery` and `startingPosition`, depending on `initialPosition`.

```
if (initialPosition == TRIM_HORIZON) {
  discoverAllShards()
  startingPosition == TRIM_HORIZON
} else if (initialPosition == LATEST) {
  # Initial discovery
  discoverAllShards()
  if shard is open,
     consume from shard
     startingPosition == LATEST
  if shard is closed,
     mark shard as FINISHED
  
  # Periodic discovery
  discoverAllShards()
  if shard already known, skip
  if shard is not known, startingPosition == LATEST
  
}
```

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
```diff
@@ -0,0 +1,237 @@
...
+     * Mark splits as assigned. Assigned splits will no longer be returned as pending splits.
+     *
+     * @param splitsToAssign collection of splits to mark as assigned
+     */
+    public void markAsAssigned(Collection<DynamoDbStreamsShardSplit> splitsToAssign) {
+        splitsToAssign.forEach(split -> assignedSplits.add(split.splitId()));
+    }
+
+    /**
+     * Mark splits as finished. Assigned splits will no longer be returned as pending splits.
```
hlteoh37, line 137: nit: this should be `Finished`

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
```diff
@@ -0,0 +1,322 @@
...
+     */
+    public List<DynamoDbStreamsShardSplit> splitsAvailableForAssignment() {
+        return knownSplits.values().stream()
+                .filter(
+                        split -> {
+                            boolean splitIsNotAssigned = !isAssigned(split.splitId());
+                            return splitIsNotAssigned
+                                    && !isFinished(split.splitId())
+                                    && (verifyAllParentSplitsAreFinished(split)
+                                            || verifyAllParentSplitsAreCleanedUp(split));
```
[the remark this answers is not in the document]
hlteoh37: Unresolving as this has not been pushed down! Given we have only 1 parent - we can also simplify this logic

hlteoh37 (changes requested):
Added additional comments

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/proxy/DynamoDbStreamsProxy.java
```diff
@@ -86,13 +88,32 @@ public GetRecordsResponse getRecords(
             }
             return getRecordsResponse;
         } catch (ExpiredIteratorException e) {
+            LOG.warn(
+                    "Received ExpiredIteratorException from GetRecords. "
+                            + "Calling GetShardIterator for shard: {} with position: {}",
+                    startingPosition,
+                    shardId);
```
hlteoh37, line 95: nit: This should be something like `info`, since we automatically recover here. This might happen alot of the job is backpressured.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/proxy/DynamoDbStreamsProxy.java
```diff
@@ -168,8 +194,15 @@ private String getShardIterator(
         try {
             return dynamoDbStreamsClient.getShardIterator(requestBuilder.build()).shardIterator();
         } catch (ResourceNotFoundException e) {
-            LOG.info(
-                    "Received ResourceNotFoundException. "
+            LOG.warn(
+                    "Received ResourceNotFoundException from GetShardIterator. "
+                            + "Shard {} of stream {} is no longer valid, marking it as complete.",
```
hlteoh37, line 199: this is good to be warn. Good call.

Let's add a warning here too that this might indicate data loss (since shard was not fully read, but it already expired)

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/proxy/DynamoDbStreamsProxy.java
```diff
@@ -127,8 +153,8 @@ private DescribeStreamResponse describeStream(String streamArn, @Nullable String
                 dynamoDbStreamsClient.describeStream(describeStreamRequest);
 
         StreamStatus streamStatus = describeStreamResponse.streamDescription().streamStatus();
-        if (!(streamStatus.equals(StreamStatus.DISABLED)
-                || streamStatus.equals(StreamStatus.ENABLING.toString()))) {
+        if (streamStatus.equals(StreamStatus.DISABLED)
```
hlteoh37, line 156: As discussed offline - this should only be `ENABLING`

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/PollingDynamoDbStreamsShardSplitReader.java
```diff
@@ -75,11 +82,17 @@ public RecordsWithSplitIds<Record> fetch() throws IOException {
...
+        try {
+            getRecordsResponse =
+                    dynamodbStreams.getRecords(
+                            splitState.getStreamArn(),
+                            splitState.getShardId(),
+                            splitState.getNextStartingPosition());
+        } catch (ResourceNotFoundException e) {
+            return new DynamoDbStreamRecordsWithSplitIds(
+                    Collections.emptyIterator(), splitState.getSplitId(), true);
+        }
```
hlteoh37, line 95: nit: I think this is not needed anymore, because we handle it inside the streamProxy.  No harm keeping though. We should add a WARN log

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/split/DynamoDbStreamsShardSplitSerializer.java
```diff
@@ -63,6 +65,11 @@ public byte[] serialize(DynamoDbStreamsShardSplit split) throws IOException {
                     out.writeUTF((String) startingMarker);
                 }
             }
+            out.writeInt(split.getParentShardIds().size());
+            for (String parentShardId : split.getParentShardIds()) {
+                out.writeUTF(parentShardId);
+            }
```
hlteoh37, line 71: can be simplified once we remove the set of `parentShardIds`

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
```diff
@@ -0,0 +1,237 @@
...
+                knownSplits.remove(finishedSplitId);
+            }
+        }
+    }
+
+    private boolean isSplitReadyToBeCleanedUp(
+            DynamoDbStreamsShardSplit finishedSplit, Set<String> discoveredSplitIds) {
+        String splitId = finishedSplit.splitId();
+        boolean parentSplitsAreFinished = verifyAllParentSplitsAreFinished(finishedSplit);
+        boolean parentSplitsAreCleanedUp = verifyAllParentSplitsAreCleanedUp(finishedSplit);
```
hlteoh37, line 212: why do we care about parents? Might cause issues when we restore from stale snapshot (all shards are expired and should be cleaned up)

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/config/DynamodbStreamsSourceConfigConstants.java
```diff
@@ -40,9 +42,41 @@ public enum InitialPosition {
...
+    public static final ConfigOption<Duration> DYNAMODB_STREAMS_EXPONENTIAL_BACKOFF_MAX_DELAY =
+            ConfigOptions.key("flink.dynamodbstreams.backoff.maxdelay")
+                    .durationType()
+                    .defaultValue(Duration.ofMillis(1000))
+                    .withDescription(
+                            "The maximum delay for exponential backoff for describestream");
+
+    public static final Duration DDB_STREAMS_MAX_RETENTION_PERIOD = Duration.ofHours(24);
+
+    public static final String SHARD_ID_SEPARATOR = "-";
```
hlteoh37, line 78: These are not strictly `Config`, so putting it here in public makes it public interface. Let's make it private in the class that uses them

hlteoh37: Also ! I wonder if we should add 1 hour here to reduce race conditions?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/config/DynamodbStreamsSourceConfigConstants.java
```diff
@@ -40,9 +42,41 @@ public enum InitialPosition {
     public static final ConfigOption<Long> SHARD_DISCOVERY_INTERVAL_MILLIS =
             ConfigOptions.key("flink.shard.discovery.intervalmillis")
                     .longType()
-                    .defaultValue(10000L)
+                    .defaultValue(60000L)
                     .withDescription("The interval between each attempt to discover new shards.");
```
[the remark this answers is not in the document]
gguptp: changed it to duration type

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/config/DynamodbStreamsSourceConfigConstants.java
gguptp: yeah, having 25 hours would reduce the race conditions. Will change it. Sure, will make these private in the class which uses them

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
```diff
@@ -87,25 +92,29 @@ public DynamoDbStreamsSourceEnumerator(
...
     }
 
     @Override
     public void start() {
-        if (lastSeenShardId == null) {
-            context.callAsync(this::initialDiscoverSplits, this::assignSplits);
+        // if there is already a saved state, we can just start with that state instead of waiting
+        // for shard discovery to complete
+        if (!hasInitialState) {
+            context.callAsync(this::discoverSplits, this::processDiscoveredSplits);
```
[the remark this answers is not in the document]
gguptp: there might be cases where parent is finished, the job died and children are unassigned. Its an edge case though, so thought its better to handle here

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
```diff
@@ -116,24 +118,96 @@ public void handleSplitRequest(int subtaskId, @Nullable String requesterHostname
...
+                    "There are inconsistencies in DescribeStream which we were not able to resolve. First leaf node on which inconsistency was detected:"
+                            + splitGraphInconsistencyTracker.getEarliestClosedLeafNode());
+            return;
+        }
+
+        splitTracker.addSplits(splitGraphInconsistencyTracker.getNodes());
+        splitTracker.removeSplits(
+                splitGraphInconsistencyTracker.getNodes().stream()
+                        .map(Shard::shardId)
+                        .collect(Collectors.toSet()));
```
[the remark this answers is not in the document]
gguptp: renamed it to `cleanUpOldFinishedSplits`

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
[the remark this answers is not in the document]
gguptp: yeah, agreed, moving this to discoverSplits()

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
```diff
@@ -116,24 +118,96 @@ public void handleSplitRequest(int subtaskId, @Nullable String requesterHostname
...
+        SplitGraphInconsistencyTracker splitGraphInconsistencyTracker =
+                new SplitGraphInconsistencyTracker();
+        splitGraphInconsistencyTracker.addNodes(discoveredSplits.getShards());
+
+        boolean streamDisabled = discoveredSplits.getStreamStatus().equals(StreamStatus.DISABLED);
+        int describeStreamInconsistencyResolutionCount =
+                sourceConfig.get(DESCRIBE_STREAM_INCONSISTENCY_RESOLUTION_RETRY_COUNT);
+        for (int i = 0;
+                i < describeStreamInconsistencyResolutionCount
+                        && !streamDisabled
```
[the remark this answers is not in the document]
gguptp: added a comment

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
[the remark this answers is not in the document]
gguptp: made it to warn

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
gguptp: we discussed this offline, theres not really a good way to detect a shard whose parent has expired during snapshot restore from a shard which is a separate lineage, hence we prefer simplicity here, will add comment in the code regarding the same

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
gguptp: yeah, will remove the set

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
gguptp: we only want to clean a split up if its parents are read or have been cleaned up. For a restore from stale snapshot, we'll be assigning the splits and GetRecords or GetShardIterator will mark the splits as finished and they'll get put in finished splits.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/proxy/DynamoDbStreamsProxy.java
gguptp: fixed

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
gguptp: removed set, made it as a String

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
gguptp: updated

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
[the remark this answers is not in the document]
gguptp: simplifed this logic

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/proxy/DynamoDbStreamsProxy.java
gguptp: changed this to info

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/proxy/DynamoDbStreamsProxy.java
```diff
@@ -102,18 +123,23 @@ public void close() throws IOException {
         httpClient.close();
     }
 
-    private List<Shard> getShardsOfStream(String streamName, @Nullable String lastSeenShardId) {
-        List<Shard> shardsOfStream = new ArrayList<>();
+    private ListShardsResult getShardsOfStream(
```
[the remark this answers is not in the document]
gguptp: yeah i don't think having this function makes sense. Moved it to ListShards

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/proxy/DynamoDbStreamsProxy.java
```diff
@@ -102,18 +123,23 @@ public void close() throws IOException {
...
-            List<Shard> shards = describeStreamResponse.streamDescription().shards();
-            shardsOfStream.addAll(shards);
-
+            describeStreamResponse = this.describeStream(streamName, lastEvaluatedShardId);
+            listShardsResult.addShards(describeStreamResponse.streamDescription().shards());
+            listShardsResult.setStreamStatus(
+                    describeStreamResponse.streamDescription().streamStatus());
+            lastEvaluatedShardId =
+                    describeStreamResponse.streamDescription().lastEvaluatedShardId();
+            LOG.debug("DescribeStream lastEvaluatedShardId: {}", lastEvaluatedShardId);
```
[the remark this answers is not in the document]
gguptp: added a debug log for full list of shards

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/proxy/DynamoDbStreamsProxy.java
gguptp: added a warning in the log

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/PollingDynamoDbStreamsShardSplitReader.java
gguptp: i don't think getRecords in streamProxy handles this.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/split/DynamoDbStreamsShardSplitSerializer.java
gguptp: changed

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
hlteoh37: Followup JIRA https://issues.apache.org/jira/browse/FLINK-36204

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/PollingDynamoDbStreamsShardSplitReader.java
```diff
@@ -75,7 +81,8 @@ public RecordsWithSplitIds<Record> fetch() throws IOException {
             return INCOMPLETE_SHARD_EMPTY_RECORDS;
         }
 
-        GetRecordsResponse getRecordsResponse =
+        GetRecordsResponse getRecordsResponse;
+        getRecordsResponse =
```
hlteoh37, line 85: nit: can be merged

hlteoh37 (changes requested):
LGTM - one more small change and we are good to go!

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/PollingDynamoDbStreamsShardSplitReader.java
gguptp: done

hlteoh37:
Validated that unit test coverage is pretty decent 

image

PR #152, opened 2024-07-24
[FLINK-35500][Connectors/DynamoDB] DynamoDb Table API Sink fails to delete elements due to key not found

robg-eb:
## Purpose of the change

When `DynamoDbSink` is used with CDC sources, it fails to process DELETE records and throws
```
    org.apache.flink.connector.dynamodb.shaded.software.amazon.awssdk.services.dynamodb.model.DynamoDbException: The provided key element does not match the schema
```

This is due to `DynamoDbSinkWriter` passing the whole DynamoDb Item as key instead of the constructed primary key alone.

## Verifying this change
- Added tests to `RowDataToAttributeValueConverterTest.java`:
  - `testDeleteOnlyPrimaryKey` - Ensures that for a `DELETE` request, only the (single) PK field is included
  - `testDeleteOnlyPrimaryKeys` - Ensures that for a `DELETE` request with a composite PK, both PK fields are included.
  - `testPKIgnoredForInsert` - Ensures that PK is ignored when an `INSERT` request is done, and all fields continue to be included as they have been in the past.
  - `testPKIgnoredForUpdateAfter` - Ensures that PK is ignored when an `UPDATE_AFTER` request is done, and all fields continue to be included as they have been in the past.

- Ran manual tests following the steps noted in https://issues.apache.org/jira/browse/FLINK-35500 under "Steps To Reproduce".   Running the SQL statement as described in Step 6 now properly runs a DELETE in DynamoDB.

 
## Significant changes
Previously, the `PRIMARY KEY` field had no significance for a DynamoDB Sink via Table API.   Now, the `PRIMARY KEY` is required when processing a CDC stream that contains `DELETES`.    This is not a 'breaking change' because the previous behavior for processing a CDC stream containing `DELETES` was already a failure (`The provided key element does not match the schema`).  This change now provides a clear exception informing users to specify a Primary Key to avoid that failure.   To clarify this change, the PR contains updates to the Connector documentation.

### DIFF
--- a/docs/content/docs/connectors/table/dynamodb.md
+++ b/docs/content/docs/connectors/table/dynamodb.md
@@ -303,6 +303,46 @@ WITH (
 );
 ```
 
+## Specifying Primary Key For Deletes
+
+The DynamoDB sink supports Delete requests to DynamoDB, but AWS requires that a Dynamo Delete request contain **only** the key field(s), or else the Delete request will fail with `DynamoDbException: The provided key element does not match the schema`.
+Thus, if a Changelog stream being is being written to DynamoDB that contains DELETEs, you must specify the `PRIMARY KEY` on the table.
+This `PRIMARY KEY` specified for the Flink SQL Table must match the actual Primary Key of the DynamoDB table - so it must be either just the Partition Key, or in case of a composite primary key, it must be the Partition Key and Sort Key.
+
+Example For Partition Key as only Primary Key:
+```sql
+CREATE TABLE DynamoDbTable (
+  `user_id` BIGINT,
+  `item_id` BIGINT,
+  `category_id` BIGINT,
+  `behavior` STRING,
+  PRIMARY KEY (user_id) NOT ENFORCED
+) 
+WITH (
+  'connector' = 'dynamodb',
+  'table-name' = 'user_behavior',
+  'aws.region' = 'us-east-2'
+);
+```
+
+Example For Partition Key and Sort Key as Composite Primary Key:
+```sql
+CREATE TABLE DynamoDbTable (
+  `user_id` BIGINT,
+  `item_id` BIGINT,
+  `category_id` BIGINT,
+  `behavior` STRING,
+  PRIMARY KEY (user_id, item_id) NOT ENFORCED
+) 
+WITH (
+  'connector' = 'dynamodb',
+  'table-name' = 'user_behavior',
+  'aws.region' = 'us-east-2'
+);
+```
+
+Note that this Primary Key functionality, specified by `PRIMARY KEY`, can be used alongside the Sink Partitioning mentioned above via `PARTITIONED BY` to dedeuplicate data and support DELETEs.
+
 ## Notice
 
 The current implementation of the DynamoDB SQL connector is write-only and doesn't provide an implementation for source queries.
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSink.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSink.java
@@ -51,6 +51,7 @@ public class DynamoDbDynamicSink extends AsyncDynamicTableSink<DynamoDbWriteRequ
     private final Properties dynamoDbClientProperties;
     private final DataType physicalDataType;
     private final Set<String> overwriteByPartitionKeys;
+    private final Set<String> primaryKeys;
 
     protected DynamoDbDynamicSink(
             @Nullable Integer maxBatchSize,
@@ -62,7 +63,8 @@ public class DynamoDbDynamicSink extends AsyncDynamicTableSink<DynamoDbWriteRequ
             boolean failOnError,
             Properties dynamoDbClientProperties,
             DataType physicalDataType,
-            Set<String> overwriteByPartitionKeys) {
+            Set<String> overwriteByPartitionKeys,
+            Set<String> primaryKeys) {
         super(
                 maxBatchSize,
                 maxInFlightRequests,
@@ -74,6 +76,7 @@ public class DynamoDbDynamicSink extends AsyncDynamicTableSink<DynamoDbWriteRequ
         this.dynamoDbClientProperties = dynamoDbClientProperties;
         this.physicalDataType = physicalDataType;
         this.overwriteByPartitionKeys = overwriteByPartitionKeys;
+        this.primaryKeys = primaryKeys;
     }
 
     @Override
@@ -89,7 +92,7 @@ public class DynamoDbDynamicSink extends AsyncDynamicTableSink<DynamoDbWriteRequ
                         .setFailOnError(failOnError)
                         .setOverwriteByPartitionKeys(new ArrayList<>(overwriteByPartitionKeys))
                         .setDynamoDbProperties(dynamoDbClientProperties)
-                        .setElementConverter(new RowDataElementConverter(physicalDataType));
+                        .setElementConverter(new RowDataElementConverter(physicalDataType, primaryKeys));
 
         addAsyncOptionsToSinkBuilder(builder);
 
@@ -108,7 +111,8 @@ public class DynamoDbDynamicSink extends AsyncDynamicTableSink<DynamoDbWriteRequ
                 failOnError,
                 dynamoDbClientProperties,
                 physicalDataType,
-                overwriteByPartitionKeys);
+                overwriteByPartitionKeys,
+                primaryKeys);
     }
 
     @Override
@@ -136,6 +140,7 @@ public class DynamoDbDynamicSink extends AsyncDynamicTableSink<DynamoDbWriteRequ
         private Properties dynamoDbClientProperties;
         private DataType physicalDataType;
         private Set<String> overwriteByPartitionKeys;
+        private Set<String> primaryKeys;
 
         public DynamoDbDynamicTableSinkBuilder setTableName(String tableName) {
             this.tableName = tableName;
@@ -164,6 +169,11 @@ public class DynamoDbDynamicSink extends AsyncDynamicTableSink<DynamoDbWriteRequ
             return this;
         }
 
+        public DynamoDbDynamicTableSinkBuilder setPrimaryKeys(Set<String> primaryKeys) {
+            this.primaryKeys = primaryKeys;
+            return this;
+        }
+
         @Override
         public AsyncDynamicTableSink<DynamoDbWriteRequest> build() {
             return new DynamoDbDynamicSink(
@@ -176,7 +186,8 @@ public class DynamoDbDynamicSink extends AsyncDynamicTableSink<DynamoDbWriteRequ
                     failOnError,
                     dynamoDbClientProperties,
                     physicalDataType,
-                    overwriteByPartitionKeys);
+                    overwriteByPartitionKeys,
+                    primaryKeys);
         }
     }
 }
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSink.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSink.java
@@ -92,7 +92,8 @@ public class DynamoDbDynamicSink extends AsyncDynamicTableSink<DynamoDbWriteRequ
                         .setFailOnError(failOnError)
                         .setOverwriteByPartitionKeys(new ArrayList<>(overwriteByPartitionKeys))
                         .setDynamoDbProperties(dynamoDbClientProperties)
-                        .setElementConverter(new RowDataElementConverter(physicalDataType, primaryKeys));
+                        .setElementConverter(
+                                new RowDataElementConverter(physicalDataType, primaryKeys));
 
         addAsyncOptionsToSinkBuilder(builder);
 
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
@@ -58,8 +58,15 @@ public class DynamoDbDynamicSinkFactory extends AsyncDynamicTableSinkFactory {
                         .setDynamoDbClientProperties(
                                 dynamoDbConfiguration.getSinkClientProperties());
 
-        if (catalogTable.getResolvedSchema().getPrimaryKey().isPresent()){
-            builder = builder.setPrimaryKeys(new HashSet<>(catalogTable.getResolvedSchema().getPrimaryKey().get().getColumns()));
+        if (catalogTable.getResolvedSchema().getPrimaryKey().isPresent()) {
+            builder =
+                    builder.setPrimaryKeys(
+                            new HashSet<>(
+                                    catalogTable
+                                            .getResolvedSchema()
+                                            .getPrimaryKey()
+                                            .get()
+                                            .getColumns()));
         }
 
         addAsyncOptionsToBuilder(dynamoDbConfiguration.getAsyncSinkProperties(), builder);
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
@@ -58,6 +58,10 @@ public class DynamoDbDynamicSinkFactory extends AsyncDynamicTableSinkFactory {
                         .setDynamoDbClientProperties(
                                 dynamoDbConfiguration.getSinkClientProperties());
 
+        if (catalogTable.getResolvedSchema().getPrimaryKey().isPresent()){
+            builder = builder.setPrimaryKeys(new HashSet<>(catalogTable.getResolvedSchema().getPrimaryKey().get().getColumns()));
+        }
+
         addAsyncOptionsToBuilder(dynamoDbConfiguration.getAsyncSinkProperties(), builder);
 
         return builder.build();
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverter.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverter.java
@@ -21,10 +21,12 @@ package org.apache.flink.connector.dynamodb.table;
 import org.apache.flink.annotation.Internal;
 import org.apache.flink.connector.dynamodb.table.converter.ArrayAttributeConverterProvider;
 import org.apache.flink.table.api.DataTypes;
+import org.apache.flink.table.api.TableException;
 import org.apache.flink.table.data.RowData;
 import org.apache.flink.table.data.conversion.DataStructureConverters;
 import org.apache.flink.table.types.DataType;
 import org.apache.flink.table.types.KeyValueDataType;
+import org.apache.flink.types.RowKind;
 
 import software.amazon.awssdk.enhanced.dynamodb.AttributeConverterProvider;
 import software.amazon.awssdk.enhanced.dynamodb.EnhancedType;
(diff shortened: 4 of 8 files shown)


flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverterTest.java
```diff
@@ -547,12 +552,132 @@ void testInstantArray() {
         assertThat(actualResult).containsAllEntriesOf(expectedResult);
     }
 
+    @Test
```
vahmed-hamdy, line 555: Could we add tests for delete with no primary key and verify exception

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverter.java
```diff
@@ -43,14 +47,35 @@ public class RowDataToAttributeValueConverter {
 
     private final DataType physicalDataType;
     private final TableSchema<RowData> tableSchema;
+    private final Set<String> primaryKeys;
 
-    public RowDataToAttributeValueConverter(DataType physicalDataType) {
+    public RowDataToAttributeValueConverter(DataType physicalDataType, Set<String> primaryKeys) {
```
vahmed-hamdy, line 52: nit: could we overload the constructor instead and pass empty set on default?

vahmed-hamdy (changes requested):

vahmed-hamdy:
There are some style violations, Could you run `mvn spotless:apply` ahead of the PR please?

robg-eb:
@vahmed-hamdy - 

> There are some style violations, Could you run `mvn spotless:apply` ahead of the PR please?

Done, I've applied updated formatting now!

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverterTest.java
```diff
@@ -547,12 +552,132 @@ void testInstantArray() {
...
+        RowData rowData = createElementWithMultipleFields(StringData.fromString(value), StringData.fromString(otherValue));
+        rowData.setRowKind(RowKind.INSERT);
+
+        Map<String, AttributeValue> actualResult =
+                rowDataToAttributeValueConverter.convertRowData(rowData);
+        Map<String, AttributeValue> expectedResult = new HashMap<>();
+        expectedResult.put(key, AttributeValue.builder().s(value).build());
+        expectedResult.put(otherField, AttributeValue.builder().s(otherValue).build());
+
+        assertThat(actualResult).containsAllEntriesOf(expectedResult);
```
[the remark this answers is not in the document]
robg-eb: @vahmed-hamdy - I took a look at doing this, but `containsExactly` only works for (ordered) Lists, whereas the things we're comparing here are actually `HashMap` .   We'd have to convert these HashMap and impose an order on them as lists to compare them, which would seem to be more complicated than what I have here in just running two separate assertions - would you agree ?  Let me know if you think there's a better way - I'm admittedly not a Java expert..

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverterTest.java
```diff
@@ -547,12 +552,132 @@ void testInstantArray() {
...
+                new RowDataToAttributeValueConverter(dataType, primaryKeys);
+        RowData rowData = createElementWithMultipleFields(StringData.fromString(value), StringData.fromString(otherValue));
+        rowData.setRowKind(RowKind.DELETE);
+
+        Map<String, AttributeValue> actualResult =
+                rowDataToAttributeValueConverter.convertRowData(rowData);
+        Map<String, AttributeValue> expectedResult =
+                singletonMap(key, AttributeValue.builder().s(value).build());
+
+        assertThat(actualResult).containsAllEntriesOf(expectedResult);
```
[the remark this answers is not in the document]
robg-eb: See response below

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverterTest.java
```diff
@@ -547,12 +552,132 @@ void testInstantArray() {
...
+                StringData.fromString(value), StringData.fromString(additionalValue), StringData.fromString(otherValue));
+        rowData.setRowKind(RowKind.DELETE);
+
+        Map<String, AttributeValue> actualResult =
+                rowDataToAttributeValueConverter.convertRowData(rowData);
+        Map<String, AttributeValue> expectedResult = new HashMap<>();
+        expectedResult.put(key, AttributeValue.builder().s(value).build());
+        expectedResult.put(additionalKey, AttributeValue.builder().s(additionalValue).build());
+
+        assertThat(actualResult).containsAllEntriesOf(expectedResult);
```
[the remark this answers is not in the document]
robg-eb: See response below

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverterTest.java
robg-eb: Yes, great idea, I've done this now in latest commits!

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverter.java
z3d1k: Instead of duplicating logic, it will be better to call constructor with all parameters from one with defaults. This way there only 1 place to change if required:
```
    public RowDataToAttributeValueConverter(DataType physicalDataType) {
        this(physicalDataType, Collections.emptySet());
    }
```

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/RowDataElementConverterTest.java
```diff
@@ -45,10 +48,17 @@ public class RowDataElementConverterTest {
...
     private static final RowDataElementConverter elementConverter =
-            new RowDataElementConverter(DATA_TYPE);
+            new RowDataElementConverter(DATA_TYPE, null);
     private static final SinkWriter.Context context = new UnusedSinkWriterContext();
     private static final RowDataToAttributeValueConverter rowDataToAttributeValueConverter =
-            new RowDataToAttributeValueConverter(DATA_TYPE);
+            new RowDataToAttributeValueConverter(DATA_TYPE, null);
+
+    private static final Set<String> primaryKeys =
+            new HashSet<>(Collections.singletonList("partition_key"));
```
z3d1k, line 57: nit: You can use `Collections.singleton(...)` to create Set (immutable) with a single element.

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverterTest.java
```diff
@@ -51,7 +58,7 @@ void testChar() {
 
         DataType dataType = DataTypes.ROW(DataTypes.FIELD(key, DataTypes.CHAR(9)));
         RowDataToAttributeValueConverter rowDataToAttributeValueConverter =
-                new RowDataToAttributeValueConverter(dataType);
+                new RowDataToAttributeValueConverter(dataType, null);
```
z3d1k, line 61: Here and below, you can remove `null` argument since you've added constructor overload

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverterTest.java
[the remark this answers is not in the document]
z3d1k: You can use `assertThat(...).containsExactlyInAnyOrderEntriesOf(...)` here to check that maps contain same elements without worrying about order.

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverterTest.java
```diff
@@ -547,12 +554,180 @@ void testInstantArray() {
...
+        assertThat(expectedResult).containsAllEntriesOf(actualResult);
+    }
+
+    @Test
+    void testPKIgnoredForUpdateAfter() {
+        String key = "key";
+        String value = "some_string";
+        String otherField = "other_field";
+        String otherValue = "other_value";
+        Set<String> primaryKeys = new HashSet<>(Collections.singletonList(key));
```
z3d1k, line 691: nit: `Collections.singleton(key)`

z3d1k (changes requested):
Thank you for the contribution!
Posted a few notes on code, but overall implementation looks good

@hlteoh37 please take a look

vahmed-hamdy:
Thanks @robg-eb , 
@hlteoh37 could you please take a look

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/RowDataElementConverterTest.java
robg-eb: Implemented these changes now.

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverterTest.java
robg-eb: Thanks!  Implemented that change.

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverterTest.java
robg-eb: Thanks, addressed this now.

robg-eb:
I addressed all the open comments from @z3d1k now.  Thank you all for the suggestions!   Sounds like @hlteoh37 is the next who should take a look?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSink.java
```diff
@@ -62,7 +63,8 @@ protected DynamoDbDynamicSink(
             boolean failOnError,
             Properties dynamoDbClientProperties,
             DataType physicalDataType,
-            Set<String> overwriteByPartitionKeys) {
+            Set<String> overwriteByPartitionKeys,
+            Set<String> primaryKeys) {
```
dzikosc, line 67: In DynamoDB nomenclature, this set of properties is a primaryKey (not primaryKeys). Let's use a proper name

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
```diff
@@ -58,6 +58,17 @@ public DynamicTableSink createDynamicTableSink(Context context) {
                         .setDynamoDbClientProperties(
                                 dynamoDbConfiguration.getSinkClientProperties());
 
+        if (catalogTable.getResolvedSchema().getPrimaryKey().isPresent()) {
+            builder =
+                    builder.setPrimaryKeys(
+                            new HashSet<>(
```
dzikosc, line 64: DynamoDB primary key is an ordered set of properties - partitionKey and then sortKey. Let's model it like that.

I would suggest using those as dedicated properties instead in the Sink model. We could add an extra validation here to ensure consistency with DDB schema.

dzikosc (commented):
Let's model the concept of DDB primary key, using the specification and correct naming

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.CoreComponents.html#HowItWorks.CoreComponents.PrimaryKey

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
```diff
@@ -58,6 +58,17 @@ public DynamicTableSink createDynamicTableSink(Context context) {
                         .setDynamoDbClientProperties(
                                 dynamoDbConfiguration.getSinkClientProperties());
 
+        if (catalogTable.getResolvedSchema().getPrimaryKey().isPresent()) {
```
dzikosc, line 61: IN case of DynamoDB contents of PrimaryKey and Partition Key should be the same to ensure semantic correctness of various operations. 

To simplify the usage, we should aim to use just PrimaryKey property, as it's a fairly intuitive and well understood concept for DDB users. In the meantime, we should probably aim to start deprecation process of the `partition by` mechanism. 

The first step could be 
- if PK is set, but `partition by` is not, use PK as `partition by` 
- document that `partition by` is on deprecation path
- log warnings when `partition by` is being used

docs/content/docs/connectors/table/dynamodb.md
```diff
@@ -303,6 +303,46 @@ WITH (
...
+This `PRIMARY KEY` specified for the Flink SQL Table must match the actual Primary Key of the DynamoDB table - so it must be either just the Partition Key, or in case of a composite primary key, it must be the Partition Key and Sort Key.
+
+Example For Partition Key as only Primary Key:
+```sql
+CREATE TABLE DynamoDbTable (
+  `user_id` BIGINT,
+  `item_id` BIGINT,
+  `category_id` BIGINT,
+  `behavior` STRING,
+  PRIMARY KEY (user_id) NOT ENFORCED
```
dzikosc, line 319: Why do we use NOT ENFORCED qualifier? In case of DynamoDB Primary Key property (or properties) are always present on the record.

docs/content/docs/connectors/table/dynamodb.md
```diff
@@ -303,6 +303,46 @@ WITH (
...
+  PRIMARY KEY (user_id, item_id) NOT ENFORCED
+) 
+WITH (
+  'connector' = 'dynamodb',
+  'table-name' = 'user_behavior',
+  'aws.region' = 'us-east-2'
+);
+```
+
+Note that this Primary Key functionality, specified by `PRIMARY KEY`, can be used alongside the Sink Partitioning mentioned above via `PARTITIONED BY` to dedeuplicate data and support DELETEs.
```
nicusX, line 344: I would clarify that `PARTITIONED BY` is used by Flink for deduplication within the same batch (and to support delete), but it's different from DynamoDB's *partitionKey*. If the user is familiar with DynamoDB but not much with Flink this may generate lot of confusion.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSink.java
nicusX: +1

docs/content/docs/connectors/table/dynamodb.md
nicusX: Also, `PARTITIONED BY` and `PRIMARY KEY` should always be the same.
(see other comment by @dzikosc below)

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
nicusX: +1
Also, I'd add a check that, if `PARTITION BY` and `PRIMARY KEY` are both specified, they must be identical

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
nicusX: +1
It must be an ordered set/list of 1 or 2 elements.

Suggestion: if not too complicated, for a user perspective it would be clearer a no-nonsense passing a *PrimaryKey* object with two fields, *partitionKey* and *sortKey*, where *partitionKey*

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSink.java
robg-eb: To be clear, you're just suggesting changing the variable name from `primaryKeys` to `primaryKey`, right?   And leaving its data type intact of `Set<String>`, since a Primary Key can consist of two fields - the Partition Key and the Sort Key?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
robg-eb: @nicusX - While having a dedicated `PrimaryKey` object with two fields _partitionKey_ and _sortKey_ might work for the DataStream API if I were to add it to the Sink model, I am not clear on how that would then translate to the Table API.  I was hoping to just use the fact that the Table API / SQL API already supports the concept of passing in a `PRIMARY KEY` for that.  

I don't think there's any reason here why we would need to separate the partition key and sort key for the purpose here of identifying the Primary Key - and in fact, there is also some naming collision then as well as the current Table API / SQL connector supports passing in a `PARTITIONED BY` clause, adding to potential confusion.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
robg-eb: @dzikosc - Isn't there a potential use case wherein a user of the existing `PARTITION BY` functionality of the sink wants to use it to deduplicate data by something other than the actual Primary Key of the DynamoDB table?   For example, the DynamoDB table has:

```
user_id  Partition Key 
order_id Sort Key 
line_number 
price
```

Today, the user of the connector could specify a `PARTITION BY  user_id, order_id, line_number` to deduplicate incoming data - Maybe this is a contrived case, but just pointing out that we may now be blocking a use case that was previously supported.   Thoughts?  

Either way, could we keep the deprecation of `PARTITION BY` to a separate PR to reduce the scope of this one?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
nicusX: Ack for not having an object with separate fields.
However, in this case the `primaryKey` (singular) should be an ordered set (e.g. a List) and not a Set. The order of the two fields is relevant: the first always being the *partitionKey*, in DDB parlance, and the second, if present, the *sortKey*

bradodarb:
It's too bad this bugfix is withering on the vine, and really makes me wonder how many folks are actually using Dynamodb as a sink via SQL/Table API 😬. 

My team is a bit blocked by this bug, it'd be nice to pick it back up.

If core AWS maintainers aren't triaging this as a priority issue (which it should be in my unsolicited opinion 😄 )- and  @robg-eb no longer has time/energy to get this over the finish line, I'd be ok with giving it a shot.

riyarawat-amazon:
Opened https://github.com/apache/flink-connector-aws/pull/253 as a rebased continuation of this PR, addressing the outstanding review comments

RobGoretsky:
Apologies for my absence here - I left the role at the company that was using Flink and DynamoDB, so I no longer have access to an environment to test this - If someone else can carry this forward, it'd be great!  (I was also hand-writing this code in 2024, and imagine this would be way easier with an Agent's assistance now in 2026)

riyarawat-amazon:
@RobGoretsky I have opened https://github.com/apache/flink-connector-aws/pull/253 as a rebased continuation of this PR. Thanks!!

PR #156, opened 2024-08-19
[FLINK-34023][Connectors/Kinesis] Expose configuration of sink client retries

brada:
## Purpose of the change

- Expose configuration of sink client retries. 
- The consumer side already exposes client retry configuration like [flink.shard.getrecords.maxretries](https://nightlies.apache.org/flink/flink-docs-stable/api/java/org/apache/flink/streaming/connectors/kinesis/config/ConsumerConfigConstants.html#SHARD_GETRECORDS_RETRIES) but the sink producer side currently lacks similar config for PutRecords.
- The AWS SDK has deprecated RetryPolicy in favour of RetryStrategy, meaning that the old "num-retries" config is being replaced with "max-attempts" (which includes the initial attempt)
- This PR exposes the most important max-attempt config, but not the complete retry strategy, since representing all the various types of (jittered, linear, exponential etc) strategies with universal config keys is non-trivial.

## Verifying this change

- *Modified existing unit tests*
- *Added unit tests*

## Significant changes

- [X] New feature has been introduced
  - If yes, how is this documented? (not applicable / docs / JavaDocs / not documented): **not documented**

### DIFF
--- a/flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/util/AWSClientUtil.java
+++ b/flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/util/AWSClientUtil.java
@@ -25,6 +25,7 @@ import org.apache.flink.runtime.util.EnvironmentInformation;
 import software.amazon.awssdk.awscore.client.builder.AwsAsyncClientBuilder;
 import software.amazon.awssdk.awscore.client.builder.AwsClientBuilder;
 import software.amazon.awssdk.awscore.client.builder.AwsSyncClientBuilder;
+import software.amazon.awssdk.awscore.retry.AwsRetryStrategy;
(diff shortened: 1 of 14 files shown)


brada:
It look like this PR might not be necessary if [https://github.com/apache/flink-connector-aws/pull/150](https://github.com/apache/flink-connector-aws/pull/150) gets merged.
That one mentions needing to upgrade the AWS SDK for the new RetryStrategy class, which I think explains why the e2e tests are failing here (https://lists.apache.org/thread/cotopq8qs2mk4cysqd2jk3jvr6lhj6cd)

hlteoh37:
>It look like this PR might not be necessary if https://github.com/apache/flink-connector-aws/pull/150 gets merged.

Yep we are working on merging that instead. Thanks @brada !

PR #158, opened 2024-09-04
[FLINK-36215][dynamodb][BugFix] Removing dependency on sorting of splits

gguptp:
…lits

## Purpose of the change

Removing dependency on sorting of splits. This also makes sure that if parent and child come in same describestream call, we get TRIM_HORIZON for the child instead of race conditions where parent and child both get assigned LATEST shard iterator

## Verifying this change

This change is already covered by existing tests,

## Significant changes

### DIFF
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitGraphInconsistencyTracker.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitGraphInconsistencyTracker.java
@@ -49,9 +49,25 @@ public class SplitGraphInconsistencyTracker {
         closedLeafNodes = new TreeSet<>();
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


hlteoh37:
Let's create a new JIRA for this

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitGraphInconsistencyTracker.java
```diff
@@ -53,16 +53,22 @@ public void addNodes(List<Shard> shards) {
         for (Shard shard : shards) {
             addNode(shard);
         }
+        for (Shard shard : shards) {
```
hlteoh37, line 56: Let's add a comment here why we need 2x loops

gguptp: makes sense, updated

PR #160, opened 2024-09-04
[FLINK-36187][Connectors/Kinesis] Implement UniformShardAssigner usin…

hlteoh37:
…g Kinesis shard hashes in the KinesisStreamsSource

## Purpose of the change
Reimplement `UniformShardAssigner` to use hash key ranges instead of using round-robin. This means that the shard distribution will be even during Kinesis stream resharding.

## Verifying this change
- *Added unit tests*

## Significant changes


mrapcooper:
Programmatically LGTM ... 2 checkstyle errors and some formatting issues it fix.

mrapcooper:
LGTM

PR #165, opened 2024-09-16
[FLINK-31922][Connectors/AWS] Port over Kinesis Client configurations for retry and backoff strategies

karubian:
## Purpose of the change

Port over Kinesis Client configurations for retry and backoff
Contains the changes from this PR: https://github.com/apache/flink-connector-aws/pull/150
## Verifying this change

- *Added unit tests*

## Significant changes


flink-connector-aws-base/pom.xml
```diff
@@ -87,6 +87,12 @@ under the License.
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>kinesis</artifactId>
+            <scope>test</scope>
+        </dependency>
+
```
hlteoh37, line 95: I don't think we need this anymore - lets remove it.

flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/config/AWSConfigOptions.java
```diff
@@ -0,0 +1,187 @@
...
+import static org.apache.flink.connector.aws.config.AWSConfigConstants.profileName;
+import static org.apache.flink.connector.aws.config.AWSConfigConstants.profilePath;
+import static org.apache.flink.connector.aws.config.AWSConfigConstants.roleArn;
+import static org.apache.flink.connector.aws.config.AWSConfigConstants.roleSessionName;
+import static org.apache.flink.connector.aws.config.AWSConfigConstants.roleStsEndpoint;
+import static org.apache.flink.connector.aws.config.AWSConfigConstants.secretKey;
+import static org.apache.flink.connector.aws.config.AWSConfigConstants.webIdentityTokenFile;
+
+/** Configuration options for AWS service usage. */
+public class AWSConfigOptions {
```
hlteoh37, line 38: Let's mark this a `@PublicEvolving`

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/KinesisSourceConfigOptions.java
```diff
@@ -33,28 +36,35 @@ public enum InitialPosition {
...
+                    .durationType()
+                    .defaultValue(Duration.ofSeconds(10))
                     .withDescription("The interval between each attempt to discover new shards.");
+
+    public static final ConfigOption<Integer> SHARD_GET_RECORDS_MAX =
+            ConfigOptions.key("kinesis.shard.get-records.max-record-count")
+                    .intType()
+                    .defaultValue(10000)
+                    .withDescription(
+                            "The maximum number of getRecords attempts if we get a recoverable exception.");
```
hlteoh37, line 69: Why do we need this? Also I think this description might not be correct right - this seems to be used as a "getRecords" limit.

hlteoh37 (changes requested):
Thanks for the PR @karubian. Left some comments

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/KinesisSourceConfigOptions.java
karubian: The old connector has this as configuration. [https://github.com/apache/flink-connector-aws/blob/main/flink-connector-aws/flink-[…]treaming/connectors/kinesis/config/ConsumerConfigConstants.java](https://github.com/apache/flink-connector-aws/blob/main/flink-connector-aws/flink-connector-kinesis/src/main/java/org/apache/flink/streaming/connectors/kinesis/config/ConsumerConfigConstants.java#L245) 
I will update the description.

hlteoh37 (approved):
LGTM - Thanks @karubian for the contribution!

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/KinesisSourceConfigOptions.java
```diff
@@ -19,12 +19,17 @@
...
 import org.apache.flink.configuration.ConfigOption;
 import org.apache.flink.configuration.ConfigOptions;
+import org.apache.flink.connector.aws.config.AWSConfigOptions;
+
+import java.time.Duration;
 
 /** Constants to be used with the KinesisStreamsSource. */
 @Experimental
-public class KinesisStreamsSourceConfigConstants {
+@PublicEvolving
```
hlteoh37, line 31: Hmmm, there's no point having both here - Only 1 is required. Because this connector is still `@Experimental` we can keep it that way 👀

PR #170, opened 2024-09-24
[FLINK-36357][Connectors/Kinesis] Set up AWS SDK default retry strate…

hlteoh37:
…gy configurations

## Purpose of the change

Configure SDK retry default configurations. This helps with classification of retryable errors out of the box.

## Verifying this change
- Manually run connector to verify that the application doesn't restart as frequently on connection loss!

## Significant changes

### DIFF
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
@@ -50,12 +50,10 @@ import org.apache.flink.metrics.MetricGroup;
 import org.apache.flink.util.Preconditions;
 import org.apache.flink.util.UserCodeClassLoader;
 
-import software.amazon.awssdk.awscore.exception.AwsServiceException;
-import software.amazon.awssdk.awscore.internal.AwsErrorCode;
 import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration;
+import software.amazon.awssdk.core.internal.retry.SdkDefaultRetryStrategy;
 import software.amazon.awssdk.http.SdkHttpClient;
 import software.amazon.awssdk.http.apache.ApacheHttpClient;
-import software.amazon.awssdk.retries.StandardRetryStrategy;
 import software.amazon.awssdk.retries.api.BackoffStrategy;
 import software.amazon.awssdk.retries.api.RetryStrategy;
 import software.amazon.awssdk.services.kinesis.KinesisClient;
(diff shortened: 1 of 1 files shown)


gguptp (approved):
LGTM

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
```diff
@@ -254,31 +252,10 @@ private RetryStrategy createExpBackoffRetryStrategy(
         final BackoffStrategy backoffStrategy =
                 BackoffStrategy.exponentialDelayHalfJitter(initialDelay, maxDelay);
 
-        return StandardRetryStrategy.builder()
+        return SdkDefaultRetryStrategy.standardRetryStrategyBuilder()
                 .backoffStrategy(backoffStrategy)
                 .throttlingBackoffStrategy(backoffStrategy)
                 .maxAttempts(maxAttempts)
```
karubian, line 258: Circuit breaker is enabled in this strategy by default. Is this intended?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
```diff
@@ -254,31 +252,10 @@ private RetryStrategy createExpBackoffRetryStrategy(
         final BackoffStrategy backoffStrategy =
                 BackoffStrategy.exponentialDelayHalfJitter(initialDelay, maxDelay);
 
-        return StandardRetryStrategy.builder()
+        return SdkDefaultRetryStrategy.standardRetryStrategyBuilder()
```
karubian, line 255: Do we have any thoughts on how this would work with a user defined retryable exceptions improvement like the one in the old connector?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
hlteoh37: I was wondering that - is there a reason we really don't want to enable circuit breaker for our setup?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
hlteoh37: > Do we have any thoughts on how this would work with a user defined retryable exceptions improvement like the one in the old connector?

Good qn - we can add more predicates by calling `retryOnException()`, but we won't be able to override the ones already marked as non-retryable in the SDK defaults, because I think it does eager checks for non-retryable errors.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSource.java
karubian: I don't think so. I believe it will be useful, we just need to document this as a new change as it might be surprising for users.

karubian (approved):

PR #171, opened 2024-09-24
[FLINK-31987][Connectors/AWS] Implement KDS Table API support

karubian:
## Purpose of the change

Implements the Table API for the Kinesis Source.

## Verifying this change

- *Added unit tests*

## Significant changes
- [X] Dependencies have been added or upgraded
- [X] New feature has been introduced


hlteoh37 (approved):
Thanks for fixing the E2E tests. 
We have discussed this PR extensively offline, and the changes here look good.
We will document this and explain how we manage the Table API versioning behavior.

PR #172, opened 2024-09-27
[FLINK-31986][Connectors/Kinesis] Setup integration tests for FLIP-27 Kinesis source

elphastori:
## Purpose of the change

Sets up Integrations tests for the FLIP-27 DataStreams API Kinesis source

## Verifying this change

- *Added unit tests*

### DIFF
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/pom.xml
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/pom.xml
@@ -123,6 +123,12 @@ under the License.
         </dependency>
 
         <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-jsr310</artifactId>
         </dependency>
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/src/test/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSourceITCase.java
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/src/test/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSourceITCase.java
@@ -1,0 +1,220 @@
+package org.apache.flink.connector.kinesis.source;
+
+import org.apache.flink.api.common.eventtime.WatermarkStrategy;
+import org.apache.flink.api.common.serialization.SimpleStringSchema;
+import org.apache.flink.api.common.time.Deadline;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.connector.aws.testutils.AWSServicesTestUtils;
+import org.apache.flink.connector.aws.testutils.LocalstackContainer;
+import org.apache.flink.connector.aws.util.AWSGeneralUtil;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.test.junit5.MiniClusterExtension;
+
+import com.google.common.collect.Lists;
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.AfterEach;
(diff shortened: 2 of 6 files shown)


flink-connector-aws/flink-connector-aws-kinesis-streams/pom.xml
```diff
@@ -122,6 +122,12 @@ under the License.
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3</artifactId>
+            <scope>test</scope>
+        </dependency>
+
```
hlteoh37, line 130: Why do we test using s3? Can we use kinesis instead?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/test/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSourceITCase.java
```diff
@@ -0,0 +1,218 @@
...
+                                        .withSourceConnectionStreamArn(
+                                                "arn:aws:kinesis:ap-southeast-1:000000000000:stream/stream-not-exists")
+                                        .runScenario())
+                .havingCause()
+                .havingCause()
+                .havingCause()
+                .havingCause()
+                .havingCause()
+                .havingCause()
+                .havingCause()
```
hlteoh37, line 104: Could we check this recursively instead?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/test/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSourceITCase.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    @AfterEach
+    void teardown() {
+        System.clearProperty(SdkSystemSetting.CBOR_ENABLED.property());
+        AWSGeneralUtil.closeResources(httpClient, kinesisClient);
+    }
+
+    @Test
+    void nonExistentStreamShouldResultInFailure() {
```
hlteoh37, line 89: Nice! Can we add a happy case too?

hlteoh37 (changes requested):
Nice work @elphastori . Added some comments

flink-connector-aws/flink-connector-aws-kinesis-streams/pom.xml
elphastori: We are actually testing Kinesis only. The S3 dependency is required because the LocalStack container in the connector base arbitrarily creates an S3 bucket and lists its contents to test readiness. [LocalstackContainer.java#L61-L83](https://github.com/apache/flink-connector-aws/blob/7fbca5566c02e3b2c581c9fcdb6c3cc2b633502a/flink-connector-aws-base/src/test/java/org/apache/flink/connector/aws/testutils/LocalstackContainer.java#L61-L83)

flink-connector-aws/flink-connector-aws-kinesis-streams/src/test/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSourceITCase.java
elphastori: Good point, I've simplified the assertion using `.withStackTraceContaining`. We no longer assert the exact position of the cause in the stack trace which is more robust.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/test/java/org/apache/flink/connector/kinesis/source/KinesisStreamsSourceITCase.java
elphastori: Added a happy case!

flink-connector-aws/flink-connector-aws-kinesis-streams/pom.xml
hlteoh37: Right - thanks!

hlteoh37 (approved):
LGTM - thanks @elphastori

PR #178, opened 2024-10-31
[FLINK-36637] Fix ListShards for large KDS streams

elphastori:
## Purpose of the change

Fixes ListShards for Kinesis streams with large number of shards

## Verifying this change

This change is already covered by existing unit tests in `KinesisStreamProxyTest`


hlteoh37 (approved):
LGTM. Thanks for fixing @elphastori

PR #188, opened 2025-02-01
[FLINK-32097][Connectors/Kinesis]  Implement support for Kinesis deaggregation

Lzgpom:
## Purpose of the change

This PR implements the Kinesis deaggregation. The implementation was heavily based on the old kinesis connector.

## Verifying this change

- Added integration tests for with record deaggregation
- Added unit tests
- Manually verified by running the Kinesis connector on a local Flink cluster.

## Further ToDos and Follow-ups

Checkpoints do not take into consideration the sequence number of the deaggregated records.

## Brief change log

- 767f576d4295aa6c0ad04ac34ebb183cc57bd0cd Implement record deaggregation using `AggregatorUtil` from the KCL 3.x
- ddcb1a9cf291cd9eab73829925a08da76b2dc107 Implemented tests for `RecordBatch` regarding deaggregation. There was a need for aggregated records to test. So it was used https://github.com/awslabs/kinesis-aggregation to do so. Unfortunatly, it does have a version in the maven repository compatible with KCL 3.x so it was used a workaround.
- fce6d035270509f31fd4cf3d9a13b0737cb985a1 Added integration tests.
- ddcb1a9cf291cd9eab73829925a08da76b2dc107 fce6d035270509f31fd4cf3d9a13b0737cb985a1 Added unit tests.

## Significant changes

- [x] Dependencies have been added or upgraded

### DIFF
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/pom.xml
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/pom.xml
@@ -53,6 +53,11 @@ under the License.
         </dependency>
 
         <dependency>
+            <groupId>software.amazon.kinesis</groupId>
+            <artifactId>amazon-kinesis-client</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>arns</artifactId>
         </dependency>
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/pom.xml
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/pom.xml
@@ -33,6 +33,14 @@ under the License.
     <name>Flink : Connectors : AWS : Amazon Kinesis Data Streams Connector v2</name>
     <packaging>jar</packaging>
 
+    <repositories>
+        <!-- used for the kinesis aggregator dependency since it is not available in maven central -->
+        <repository>
+            <id>jitpack.io</id>
+            <url>https://jitpack.io</url>
+        </repository>
+    </repositories>
+
     <dependencies>
         <dependency>
             <groupId>org.apache.flink</groupId>
@@ -104,6 +112,15 @@ under the License.
             <artifactId>flink-connector-aws-base</artifactId>
             <version>${project.version}</version>
             <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <!-- the kinesis aggregator dependency since it is not available in maven central -->
+            <!-- look into issue https://github.com/awslabs/kinesis-aggregation/issues/120 -->
+            <groupId>com.github.awslabs.kinesis-aggregation</groupId>
+            <artifactId>amazon-kinesis-aggregator</artifactId>
+            <version>2.0.3</version>
             <scope>test</scope>
         </dependency>
 
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/KinesisShardSplitReaderBase.java
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/KinesisShardSplitReaderBase.java
@@ -29,8 +29,8 @@ import org.apache.flink.connector.kinesis.source.split.StartingPosition;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import software.amazon.awssdk.services.kinesis.model.Record;
 import software.amazon.awssdk.services.kinesis.model.ResourceNotFoundException;
+import software.amazon.kinesis.retrieval.KinesisClientRecord;
 
 import javax.annotation.Nullable;
 
@@ -41,7 +41,6 @@ import java.util.Collections;
 import java.util.Deque;
 import java.util.HashSet;
 import java.util.Iterator;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -50,10 +49,10 @@ import static java.util.Collections.singleton;
 /** Base implementation of the SplitReader for reading from KinesisShardSplits. */
 @Internal
 public abstract class KinesisShardSplitReaderBase
-        implements SplitReader<Record, KinesisShardSplit> {
+        implements SplitReader<KinesisClientRecord, KinesisShardSplit> {
 
     private static final Logger LOG = LoggerFactory.getLogger(KinesisShardSplitReaderBase.class);
-    private static final RecordsWithSplitIds<Record> INCOMPLETE_SHARD_EMPTY_RECORDS =
+    private static final RecordsWithSplitIds<KinesisClientRecord> INCOMPLETE_SHARD_EMPTY_RECORDS =
             new KinesisRecordsWithSplitIds(Collections.emptyIterator(), null, false);
 
     private final Deque<KinesisShardSplitState> assignedSplits = new ArrayDeque<>();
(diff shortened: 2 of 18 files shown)


Lzgpom:
@hlteoh37 could you take a look?

nicusX:
I tested it locally, with a stream produced by KPL with aggregation enabled, and it worked fine.

Lzgpom:
> @Lzgpom Building this locally and I've got a spotless-check format violation. `mvn spotless:apply` fixes it

I tried to run `mvn spotless:apply` but it didn't have any changes. It allows reported that `Spotless apply skipped`

nicusX:
Uhm, `spotless:apply` it did some changes for me, and after that `mvn package` succeeded

nicusX:
If I run a simple `mvn clean package` on your branch I get spotess-check violations in `RecordBatchTest.java` + 10 other files. And  `mvn spotless:apply`  changes 12 files.
All style definitions are part of the repo.
Not sure I understand the different result

Lzgpom:
@nicusX I was using the java 17 and the spotless did not work. I changed to java 11 and it worked.
Thanks for notifying me about this.

flink-connector-aws/flink-connector-aws-kinesis-streams/pom.xml
```diff
@@ -102,6 +115,15 @@ under the License.
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <!-- the kinesis aggregator dependency since it is not available in maven central -->
+            <!-- look into issue https://github.com/awslabs/kinesis-aggregation/issues/120 -->
+            <groupId>com.github.awslabs.kinesis-aggregation</groupId>
+            <artifactId>amazon-kinesis-aggregator</artifactId>
```
nicusX, line 122: This has Apache 2.0 licence, all good so
https://github.com/awslabs/kinesis-aggregation

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/serialization/KinesisDeserializationSchema.java
```diff
@@ -60,7 +60,7 @@ default void open(DeserializationSchema.InitializationContext context) throws Ex
      * @param output the identifier of the shard the record was sent to
      * @throws IOException exception when deserializing record
      */
-    void deserialize(Record record, String stream, String shardId, Collector<T> output)
+    void deserialize(KinesisClientRecord record, String stream, String shardId, Collector<T> output)
```
nicusX, line 63: This is a breaking change to the public interface

nicusX (commented):
Changes LGTM
I verified that it now does not fail checkstyle checks.

There is a breaking change in the public API that forces a major version change.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/KinesisShardSplitReaderBase.java
```diff
@@ -50,10 +49,10 @@
 /** Base implementation of the SplitReader for reading from KinesisShardSplits. */
 @Internal
 public abstract class KinesisShardSplitReaderBase
-        implements SplitReader<Record, KinesisShardSplit> {
+        implements SplitReader<KinesisClientRecord, KinesisShardSplit> {
```
hlteoh37, line 52: This change actually changes the interface exposed for the KinesisSource, so it would be a backwards incompatible change. Is there a way we can wrap this internally?

Lzgpom: I have been looking at the implications of not changing from `Record` into `KinesisClientRecord` and this is what I have found:

- The deaggreated records would have to be converted from `KinesisClientRecord`, since `AggregatorUtil` only uses `KinesisClientRecord`,  to `Record` adding a slight overhead.
- The `KinesisClientRecord` has the `subSequenceNumber` that could and should be used for checkpointing.

Apart from that I don't see any more issues. Either way I don't mind changing it back to `Record`.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/serialization/KinesisDeserializationSchema.java
leekeiabstraction: +1

@Lzgpom Have we considered alternatives to changing the interface? I can see that Hong suggested wrapping internally. Have we considered alternatives such as having a separate KinesisClientRecordDeserializationSchema?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/KinesisShardSplitReaderBase.java
nicusX: @Lzgpom I think the gist is, can you please change the interface back to using `Record` in all `@Public` and `@PublicEvolving` public methods? 
This would allow including the change in a minor version of the connector, and users of `5.0.0` would be able to just  update the connector version with no code changes

flink-connector-aws/flink-connector-aws-kinesis-streams/pom.xml
```diff
@@ -52,6 +60,11 @@ under the License.
             <artifactId>kinesis</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>software.amazon.kinesis</groupId>
+            <artifactId>amazon-kinesis-client</artifactId>
```
dannycranmer, line 65: This is a shame, can we do this without KCL? It's a heavy weight dependency for the 1 class(?) needed for deaggregation. We have been talking about removing this dependency with the new connector for years, but never really solved this bit. Can we ask Kinesis to publish a separate library for this? Or, we could just copy that file (assuming it is still a small asking of code)

nicusX: @dannycranmer  `KinesisClientRecord` has some transitive dependencies.
However, the KCL can be substantially lightened removing most of transitive dependencies
See [this example](https://github.com/aws-samples/amazon-managed-service-for-apache-flink-examples/blob/5e4873883c64c28484261125b3a2e20df3455c84/java/KinesisSourceDeaggregation/flink-app/pom.xml#L85) which is an alternative solution, implementing de-aggregation in the Deserialization Schema

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/KinesisShardSplitReaderBase.java
Lzgpom: @nicusX I have reverted my change in the interface `KinesisDeserializationSchema` which was the only interface with `@PublicEvolving` that I changed.
In order for that to work I added a method to convert a `KinesisClientRecord` to a `Record` in the `KinesisStreamsRecordEmitter`.

Does it look good?

Lzgpom: > @Lzgpom Can you elaborate on what the user experience would be with the current changes in PR? Will it break compilation for existing apps that already depend on KinesisSource without further code changes?

@leekeiabstraction Now the changes should not break the compilation for existing apps that already depend on KinesisSource without further code changes

PR #193, opened 2025-03-25
[FLINK-37416][BugFix][Connectors/DynamoDB] Fix state inconsistency issue in DDB connector when sending split finished event from reader -> enumerator

gguptp:
## Purpose of the change

Today Flink does not support distributed consistency of events from subtask (Task Manager) to coordinator (Job Manager) - https://issues.apache.org/jira/browse/FLINK-28639. As a result we have a race condition that can lead to a shard and it's children shards stopped being processed after a job restart.
- A checkpoint started
- Enumerator took a checkpoint (shard was assigned here)
- Enumerator sent checkpoint event to reader
- Before taking reader checkpoint, a SplitFinishedEvent came up in reader
- Reader took checkpoint
- Now, just after checkpoint complete, job restarted

This can lead to a shard lineage getting lost because of a shard being in ASSIGNED state in enumerator and not being part of any task manager state.
This PR changes the behaviour by also checkpointing the finished splits events received in between two checkpoints and on restore, those events again getting replayed.

## Verifying this change

*(or)*

*(or)*

*(example:)*
- *Added unit tests*

I manually verified this by running the connector in a local flink cluster which was getting restarted every 10 minutes. I see there is no checkpoint discrepancy and there is no issue in the clusster

I also added UTs for the change

## Significant changes

### DIFF
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
@@ -139,13 +139,25 @@ public class DynamoDbStreamsSourceEnumerator
     /** When we mark a split as finished, we will only assign its child splits to the subtasks. */
     private void handleFinishedSplits(int subtaskId, SplitsFinishedEvent splitsFinishedEvent) {
         splitTracker.markAsFinished(splitsFinishedEvent.getFinishedSplitIds());
-        splitAssignment
-                .get(subtaskId)
-                .removeIf(
-                        split ->
-                                splitsFinishedEvent
-                                        .getFinishedSplitIds()
-                                        .contains(split.splitId()));
+
+        Set<DynamoDbStreamsShardSplit> splitsAssignment = splitAssignment.get(subtaskId);
+        // during recovery, splitAssignment may return null since there might be no split assigned
+        // to the subtask, but there might be SplitsFinishedEvent from that subtask.
+        // We will not do child shard assignment if that is the case since that might lead to child
+        // shards trying to get assigned before there being any readers.
+        if (splitsAssignment == null) {
+            LOG.info(
+                    "handleFinishedSplits called for subtask: {} which doesnt have any "
+                            + "assigned splits right now. This might happen due to job restarts. "
+                            + "Child shard discovery might be delayed until we have enough readers."
+                            + "Finished split ids: {}",
+                    subtaskId,
+                    splitsFinishedEvent.getFinishedSplitIds());
+            return;
+        }
+
+        splitsAssignment.removeIf(
+                split -> splitsFinishedEvent.getFinishedSplitIds().contains(split.splitId()));
         assignChildSplits(splitsFinishedEvent.getFinishedSplitIds());
     }
 
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/assigner/UniformShardAssigner.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/assigner/UniformShardAssigner.java
@@ -54,7 +54,8 @@ public class UniformShardAssigner implements DynamoDbStreamsShardAssigner {
 
         Preconditions.checkArgument(
                 selectedSubtask != -1,
-                "Expected at least one registered reader. Unable to assign split.");
+                "Expected at least one registered reader. Unable to assign split with id: %s.",
+                split.splitId());
         return selectedSubtask;
     }
 }
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
@@ -333,7 +333,13 @@ public class SplitTracker {
                 || isFinished(split.getParentShardId());
     }
 
-    private boolean isFinished(String splitId) {
+    /**
+     * Provides information whether a split is finished or not.
+     *
+     * @param splitId
+     * @return boolean value indicating if split is finished
+     */
+    public boolean isFinished(String splitId) {
         return finishedSplits.contains(splitId);
     }
 }
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReader.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReader.java
@@ -33,9 +33,14 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.services.dynamodb.model.Record;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.NavigableMap;
+import java.util.Set;
+import java.util.TreeMap;
 
 /**
  * Coordinates the reading from assigned splits. Runs on the TaskManager.
@@ -49,6 +54,8 @@ public class DynamoDbStreamsSourceReader<T>
 
     private static final Logger LOG = LoggerFactory.getLogger(DynamoDbStreamsSourceReader.class);
     private final Map<String, DynamoDbStreamsShardMetrics> shardMetricGroupMap;
+    private final NavigableMap<Long, Set<DynamoDbStreamsShardSplit>> splitFinishedEvents;
+    private long currentCheckpointId;
 
     public DynamoDbStreamsSourceReader(
             SingleThreadFetcherManager<Record, DynamoDbStreamsShardSplit> splitFetcherManager,
@@ -58,10 +65,35 @@ public class DynamoDbStreamsSourceReader<T>
             Map<String, DynamoDbStreamsShardMetrics> shardMetricGroupMap) {
         super(splitFetcherManager, recordEmitter, config, context);
         this.shardMetricGroupMap = shardMetricGroupMap;
+        this.splitFinishedEvents = new TreeMap<>();
+        this.currentCheckpointId = Long.MIN_VALUE;
     }
 
+    /**
+     * We store the finished splits in a map keyed by the checkpoint id.
+     *
+     * @param finishedSplitIds
+     */
     @Override
     protected void onSplitFinished(Map<String, DynamoDbStreamsShardSplitState> finishedSplitIds) {
+        if (finishedSplitIds.isEmpty()) {
+            return;
+        }
+
+        splitFinishedEvents.computeIfAbsent(currentCheckpointId, k -> new HashSet<>());
+        finishedSplitIds.values().stream()
+                .map(
+                        finishedSplit ->
+                                new DynamoDbStreamsShardSplit(
+                                        finishedSplit.getStreamArn(),
+                                        finishedSplit.getShardId(),
+                                        finishedSplit.getNextStartingPosition(),
+                                        finishedSplit
+                                                .getDynamoDbStreamsShardSplit()
+                                                .getParentShardId(),
+                                        true))
+                .forEach(split -> splitFinishedEvents.get(currentCheckpointId).add(split));
+
         context.sendSourceEventToCoordinator(
                 new SplitsFinishedEvent(new HashSet<>(finishedSplitIds.keySet())));
         finishedSplitIds.keySet().forEach(this::unregisterShardMetricGroup);
@@ -80,8 +112,55 @@ public class DynamoDbStreamsSourceReader<T>
 
     @Override
     public void addSplits(List<DynamoDbStreamsShardSplit> splits) {
-        splits.forEach(this::registerShardMetricGroup);
-        super.addSplits(splits);
+        List<DynamoDbStreamsShardSplit> dynamoDbStreamsShardSplits = new ArrayList<>();
+        for (DynamoDbStreamsShardSplit split : splits) {
+            if (split.isFinished()) {
+                // Replay the finished split event.
+                // We don't need to reload the split finished events in buffer back
+                // since if the next checkpoint completes, these would just be removed from the
+                // buffer. If the next checkpoint doesn't complete,
+                // we would go back to the previous checkpointed
+                // state which will again replay these split finished events.
+                context.sendSourceEventToCoordinator(
+                        new SplitsFinishedEvent(Collections.singleton(split.splitId())));
+            } else {
+                dynamoDbStreamsShardSplits.add(split);
+            }
+        }
+        dynamoDbStreamsShardSplits.forEach(this::registerShardMetricGroup);
+        super.addSplits(dynamoDbStreamsShardSplits);
+    }
+
+    /**
+     * At snapshot, we also store the pending finished split ids in the current checkpoint so that
+     * in case we have to restore the reader from state, we also send the finished split ids
+     * otherwise we run a risk of data loss during restarts of the source because of the
+     * SplitsFinishedEvent going missing.
+     *
+     * @param checkpointId
+     * @return
+     */
+    @Override
+    public List<DynamoDbStreamsShardSplit> snapshotState(long checkpointId) {
+        this.currentCheckpointId = checkpointId;
+        List<DynamoDbStreamsShardSplit> splits = new ArrayList<>(super.snapshotState(checkpointId));
+
+        if (!splitFinishedEvents.isEmpty()) {
+            // Add all finished splits to the snapshot
+            splitFinishedEvents.values().forEach(splits::addAll);
+        }
+        return splits;
+    }
+
+    /**
+     * During notifyCheckpointComplete, we should clean up the state of finished splits that are
+     * less than or equal to the checkpoint id.
+     *
+     * @param checkpointId
+     */
+    @Override
+    public void notifyCheckpointComplete(long checkpointId) {
+        splitFinishedEvents.headMap(checkpointId, true).clear();
     }
 
     @Override
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/split/DynamoDbStreamsShardSplit.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/split/DynamoDbStreamsShardSplit.java
@@ -40,12 +40,22 @@ public final class DynamoDbStreamsShardSplit implements SourceSplit {
     private final String shardId;
     private final StartingPosition startingPosition;
     private final String parentShardId;
+    private final boolean isFinished;
 
     public DynamoDbStreamsShardSplit(
             String streamArn,
             String shardId,
             StartingPosition startingPosition,
             String parentShardId) {
+        this(streamArn, shardId, startingPosition, parentShardId, false);
+    }
+
+    public DynamoDbStreamsShardSplit(
+            String streamArn,
+            String shardId,
+            StartingPosition startingPosition,
+            String parentShardId,
+            boolean isFinished) {
         checkNotNull(streamArn, "streamArn cannot be null");
         checkNotNull(shardId, "shardId cannot be null");
         checkNotNull(startingPosition, "startingPosition cannot be null");
@@ -54,6 +64,7 @@ public final class DynamoDbStreamsShardSplit implements SourceSplit {
         this.shardId = shardId;
         this.startingPosition = startingPosition;
         this.parentShardId = parentShardId;
+        this.isFinished = isFinished;
     }
 
     @Override
@@ -77,6 +88,10 @@ public final class DynamoDbStreamsShardSplit implements SourceSplit {
         return parentShardId;
     }
 
+    public boolean isFinished() {
+        return isFinished;
+    }
+
     @Override
     public String toString() {
         return "DynamoDbStreamsShardSplit{"
@@ -90,7 +105,10 @@ public final class DynamoDbStreamsShardSplit implements SourceSplit {
                 + startingPosition
                 + ", parentShardId=["
                 + parentShardId
-                + '}';
+                + "]"
+                + ", isFinished="
+                + isFinished
+                + "}";
     }
 
     @Override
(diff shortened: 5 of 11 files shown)


AHeise (commented):
Approach LGTM in general and I added some smaller nits. 

When we discussed the approach with the TreeMap, I forgot that the state of a reader is just a list of splits. So, now you used a special split to track the finished events, which is a plausible solution but also a bit awkward. (We should probably extend the interface to also allow reader-native state)

So, I provided an alternative solution where the information is layered directly inside the split and thus makes the special split unnecessary. PTAL.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
```diff
@@ -138,14 +138,26 @@ public void handleSourceEvent(int subtaskId, SourceEvent sourceEvent) {
 
     /** When we mark a split as finished, we will only assign its child splits to the subtasks. */
     private void handleFinishedSplits(int subtaskId, SplitsFinishedEvent splitsFinishedEvent) {
+        Set<DynamoDbStreamsShardSplit> splitsAssignment = splitAssignment.get(subtaskId);
```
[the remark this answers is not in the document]
gguptp: not really, can think about it as a future enhancement

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/assigner/UniformShardAssigner.java
```diff
@@ -54,7 +54,8 @@ public int assign(DynamoDbStreamsShardSplit split, Context context) {
 
         Preconditions.checkArgument(
                 selectedSubtask != -1,
-                "Expected at least one registered reader. Unable to assign split.");
+                "Expected at least one registered reader. Unable to assign split with id: "
+                        + split.splitId());
```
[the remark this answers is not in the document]
gguptp: agree, will update this

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReader.java
```diff
@@ -80,8 +105,65 @@ protected DynamoDbStreamsShardSplit toSplitType(
...
+            // Always add a dedicated split for finished state
+            splits.add(
+                    new DynamoDbStreamsShardSplit(
+                            "",
+                            DYNAMODB_STREAMS_COMPLETED_SHARD_ID,
+                            StartingPosition.fromStart(),
+                            null,
+                            splitFinishedEvents));
+        }
+        return splits;
```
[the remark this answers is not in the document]
gguptp: The only problem here is that we dont have access to the split state directly from the SourceReaderBase class and given the number of shards which DDB has, the `notifyCheckpointCompleted` function might become O(n) operation as compared to today where it is O(log(n)) operation and might hit performance

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/split/DynamoDbStreamsShardSplit.java
```diff
@@ -40,6 +43,7 @@ public final class DynamoDbStreamsShardSplit implements SourceSplit {
     private final String shardId;
     private final StartingPosition startingPosition;
     private final String parentShardId;
+    private final NavigableMap<Long, Set<String>> finishedSplitsMap;
```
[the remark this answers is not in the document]
gguptp: agree, i can just use a Set<String> here

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReaderTest.java
```diff
@@ -158,4 +165,148 @@ void testAddSplitsRegistersAndUpdatesShardMetricGroup() throws Exception {
...
+                        .filter(s -> s.getShardId().equals(split.getShardId()))
+                        .findFirst();
+
+        assertThat(originalSplitOptional)
+                .isPresent()
+                .hasValueSatisfying(
+                        originalSplit ->
+                                assertThat(originalSplit)
+                                        .usingRecursiveComparison()
+                                        .isEqualTo(split));
```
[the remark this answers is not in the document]
gguptp: agree

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReader.java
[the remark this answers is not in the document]
AHeise: The access issue can be solved by maintaining a separate `finishedSplits` collection where you add in `onSplitFinished`. `snapshotState` needs then to union the splits from SourceReaderBase and your own splits.

This would then probably also solve `notifyCheckpointCompleted` by actually reducing n to the number of finished splits. (not sure how many splits per checkpoints get finished) If you want to stick to log(N) you could combine both approaches and let `finishedSplits` be a TreeMap again. Then you really just need to add a boolean flag to the split.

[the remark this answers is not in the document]
gguptp: changed the implmentation to rather just store a finishedAfterCheckpoint in the DDB Split state. Thanks @AHeise for the suggestion! much neater

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/split/DynamoDbStreamsShardSplit.java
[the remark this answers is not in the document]
gguptp: yip, updated this, will raise a new PR revision once testing is complete

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReader.java
```diff
@@ -58,10 +65,34 @@ public DynamoDbStreamsSourceReader(
...
+     *
+     * @param finishedSplitIds
+     */
     @Override
     protected void onSplitFinished(Map<String, DynamoDbStreamsShardSplitState> finishedSplitIds) {
+        if (finishedSplitIds.isEmpty()) {
+            return;
+        }
+
+        for (Map.Entry<String, DynamoDbStreamsShardSplitState> entry :
```
tjo, line 83: We could just use `values()` instead of `entrySet()`

or as an alternative use the streaming API:
```
Set<DynamoDbStreamsShardSplit> splitFinishedEventsForCheckpoint = splitFinishedEvents.computeIfAbsent(currentCheckpointId, k -> new HashSet<>());
        finishedSplitIds.values().stream().map(entry ->
                new DynamoDbStreamsShardSplit(
                        entry.getStreamArn(),
                        entry.getShardId(),
                        entry.getNextStartingPosition(),
                        entry.getDynamoDbStreamsShardSplit().getParentShardId(),
                        currentCheckpointId)
        ).forEach(e -> splitFinishedEventsForCheckpoint.add(e));
```

gguptp: makes sense, i dont really need the entries but rather just need values

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
leekeiabstraction, line 141: Nit: Place line after `splitTracker.markAsFinished(splitsFinishedEvent.getFinishedSplitIds());` instead, moving `splitAssignment` declaration closer to where it's used.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
```diff
@@ -333,7 +333,13 @@ private boolean verifyParentIsEitherFinishedOrCleanedUp(DynamoDbStreamsShardSpli
                 || isFinished(split.getParentShardId());
     }
 
-    private boolean isFinished(String splitId) {
+    /**
+     * Provides information whether a split is finished or not.
+     *
+     * @param splitId
+     * @return
```
leekeiabstraction, line 340: Incomplete java doc. Suggest something like the below.

```
* @return      boolean value indicating if split is finished
```

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReader.java
```diff
@@ -80,8 +115,54 @@ protected DynamoDbStreamsShardSplit toSplitType(
...
+
+    /**
+     * During notifyCheckpointComplete, we should clean up the state of finished splits that are
+     * older than the checkpoint id.
+     *
+     * @param checkpointId
+     */
+    @Override
+    public void notifyCheckpointComplete(long checkpointId) {
+        splitFinishedEvents.headMap(checkpointId, true).clear();
```
leekeiabstraction, line 165: Java doc says we are cleaning up state of finished splits that are older (interpreting as exclusive) than checkpoint id, however we are providing `true` as inclusive arg in headMap. Which one is the intended behaviour?

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/split/DynamoDbStreamsShardSplit.java
```diff
@@ -40,12 +40,22 @@ public final class DynamoDbStreamsShardSplit implements SourceSplit {
     private final String shardId;
     private final StartingPosition startingPosition;
     private final String parentShardId;
+    private final long finishedAfterCheckpoint;
```
leekeiabstraction, line 43: Perhaps more of a question than comment. Would `finishedDuringCheckpoint` be more accurate? If I understand correctly, checkpointing is still in progress and the issue occurs when split happens before reader completed checkpoint.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
gguptp: makes sense

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/tracker/SplitTracker.java
gguptp: updated in next revision

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReader.java
gguptp: good point, any splits which are <= this checkpoint will be cleared. I will update the javadoc

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/split/DynamoDbStreamsShardSplit.java
gguptp: i would think finishedAfterCheckpoint is more accurate since the issue occurs when we restarted the job just after checkpoint and the reader state is up-to-date but enumerator's is not

leekeiabstraction: I agree that a restart is necessary for the reader to lose track of shards. However, the checkpoint still would cause the issue even if a restart does not occur, for example when resuming from an checkpoint/savepoint. The crux of the issue is not that shardSplit finished after checkpointing or a restart happened, but rather that shardSplit happened before reader completed checkpointing.

gguptp: actuallly i would rather say the issue exists due to following:
- Enumerator checkpointed its state
- It asked reader to checkpoint
- Just before reader started checkpointing a SplitFinished event came
- Reader checkpointed its state

Now there is a gap between enumerator and reader in terms of enumerator checkpointed first and reader checkpointed later

I guess from enumerator POV, this variable might be named as finishedAfterCheckpoint, from reader POV it can be named as finishedBeforeCheckpoint, so i'm in a dilemma what should be its name 😂

gguptp: Agreed that from the Job POV, this is during checkpoint, hence naming the variable as `finishedDuringCheckpoint`

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReader.java
```diff
@@ -80,8 +115,54 @@ protected DynamoDbStreamsShardSplit toSplitType(
...
     public void addSplits(List<DynamoDbStreamsShardSplit> splits) {
-        splits.forEach(this::registerShardMetricGroup);
-        super.addSplits(splits);
+        List<DynamoDbStreamsShardSplit> dynamoDbStreamsShardSplits = new ArrayList<>();
+        for (DynamoDbStreamsShardSplit split : splits) {
+            if (split.isFinished()) {
+                // Restore finished splits state
+                splitFinishedEvents
+                        .computeIfAbsent(split.getFinishedDuringCheckpoint(), k -> new HashSet<>())
+                        .add(split);
```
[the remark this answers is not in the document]
gguptp: I guess this might be redundant. But lets keep it as it might be easier to reason about the algo?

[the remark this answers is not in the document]
leekeiabstraction: If the part which restores the finished split is removed, there is an opportunity to simplify the Split pojo where instead of recording the checkpoint id when the split was finished, we can just have a boolean on whether split is finished.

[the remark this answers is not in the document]
gguptp: sounds reasonable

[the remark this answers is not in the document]
gguptp: updated the PR

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReader.java
```diff
@@ -58,10 +65,38 @@ public DynamoDbStreamsSourceReader(
...
+                                        finishedSplit.getShardId(),
+                                        finishedSplit.getNextStartingPosition(),
+                                        finishedSplit
+                                                .getDynamoDbStreamsShardSplit()
+                                                .getParentShardId(),
+                                        true))
+                .forEach(
+                        split ->
+                                splitFinishedEvents
+                                        .computeIfAbsent(currentCheckpointId, k -> new HashSet<>())
```
AHeise, line 97: nit you can pull that part out of the for loop.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReader.java
```diff
@@ -80,8 +115,55 @@ protected DynamoDbStreamsShardSplit toSplitType(
...
+     * @return
+     */
+    @Override
+    public List<DynamoDbStreamsShardSplit> snapshotState(long checkpointId) {
+        this.currentCheckpointId = checkpointId;
+        List<DynamoDbStreamsShardSplit> splits = super.snapshotState(checkpointId);
+
+        if (!splitFinishedEvents.isEmpty()) {
+            // Add all finished splits to the snapshot
+            splitFinishedEvents.values().forEach(splits::addAll);
```
AHeise, line 153: This implicitly assumes that the List returned from SourceReaderBase is mutable. I'd make a copy here into an arraylist to be future-proof.

AHeise (approved):
LGTM with some small nits.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
```diff
@@ -139,13 +139,25 @@ public void handleSourceEvent(int subtaskId, SourceEvent sourceEvent) {
...
-                                splitsFinishedEvent
-                                        .getFinishedSplitIds()
-                                        .contains(split.splitId()));
+
+        Set<DynamoDbStreamsShardSplit> splitsAssignment = splitAssignment.get(subtaskId);
+        // during recovery, splitAssignment may return null since there might be no split assigned
+        // to the subtask, but there might be SplitsFinishedEvent from that subtask.
+        // We will not do child shard assignment if that is the case since that might lead to child
+        // shards trying to get assigned before there being any readers.
+        if (splitsAssignment == null) {
```
hlteoh37, line 148: Why don't we make this a generic warn if we don't find the finished split in the assigned splits? (Instead of if no split is assigned to subtask)

hlteoh37 (approved):
LGTM. Added 1 nit

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
AHeise: I'd not weaken consistency checks here for all events. I'd mark the events as recovered and then allow some leniency (not even WARN because this is expected).

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReader.java
gguptp: good point

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReader.java
gguptp: agree, made the list returned by snapshotstate

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/enumerator/DynamoDbStreamsSourceEnumerator.java
gguptp: made this a info for now, will make the consistency checks stronger in a separate PR if that works?

hlteoh37: Good spot Arvid..!

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReaderTest.java
```diff
@@ -158,4 +163,128 @@ void testAddSplitsRegistersAndUpdatesShardMetricGroup() throws Exception {
...
+                .hasSize(1)
+                .allSatisfy(
+                        s -> {
+                            assertThat(s.splitId()).isEqualTo(split.splitId());
+                        });
+    }
+
+    @Test
+    void testAddSplitsWithStateRestoration() throws Exception {
+        // Create finished split with checkpoint id 1
```
leekeiabstraction, line 200: nit: Comment describes checkpoint id 1 but checkpoint id 1 is not used

gguptp: lol, good catch, fixed this

flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/source/reader/DynamoDbStreamsSourceReaderTest.java
```diff
@@ -158,4 +163,128 @@ void testAddSplitsRegistersAndUpdatesShardMetricGroup() throws Exception {
...
+        sourceReader.onSplitFinished(finishedSplits);
+
+        // Take another snapshot
+        List<DynamoDbStreamsShardSplit> snapshotSplits = sourceReader.snapshotState(2L);
+        List<DynamoDbStreamsShardSplit> snapshotFinishedSplits =
+                snapshotSplits.stream()
+                        .filter(DynamoDbStreamsShardSplit::isFinished)
+                        .collect(Collectors.toList());
+        // Verify we have 2 splits - the original split and the finished split
+        assertThat(snapshotFinishedSplits)
```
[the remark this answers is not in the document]
gguptp: fixed this

PR #197, opened 2025-04-09
[FLINK-37683][Connectors/AWS] Enable SSO credentials

soisyourface:
There is a conversation in [slack](https://apache-flink.slack.com/archives/C03G7LJTS2G/p1743713205172909) that has some additional information.
## Purpose of the change

SSO Credentials do not work without including the `sso` and `ssooidc` dependencies in the shaded jar.

## Verifying this change
I've tested this locally by doing `mvn clean install` and then using `5.1-SNAPSHOT` in my flink job. I can add further tests if this PR is wanted.

## Significant changes

### DIFF
--- a/flink-connector-aws-base/src/test/java/org/apache/flink/connector/aws/util/AWSGeneralUtilTest.java
+++ b/flink-connector-aws-base/src/test/java/org/apache/flink/connector/aws/util/AWSGeneralUtilTest.java
@@ -960,7 +960,7 @@ class AWSGeneralUtilTest {
     }
 
   @Test
-  void testGetCredentialsProviderReturnsSsoProviderForSsoProfile() throws IOException {
+  void testSsoProfileCredentialsThrowsExceptionWhenCacheFileIsMissing() throws IOException {
     String profileName = "my-sso-profile";
     String ssoStartUrl = "https://my-dummy-sso-portal.awsapps.com/start";
     String ssoRegion = "us-east-1";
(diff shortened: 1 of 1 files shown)


leekeiabstraction (changes requested):
Thank you Ryan for raising this PR!

Can we add test case? An example would be using [ProfileTokenProviderBuilder](https://github.com/aws/aws-sdk-java-v2/blob/449da7e1ec644a3f0e1b698660b10639fd62ce39/core/auth/src/main/java/software/amazon/awssdk/auth/token/credentials/ProfileTokenProvider.java#L160) to load a [test implementation of profile file](https://github.com/aws/aws-sdk-java-v2/blob/master/core/profiles/src/main/java/software/amazon/awssdk/profiles/ProfileFile.java#L61)  returns a token provider of class `SsoProfileCredentialsProvider`?

flink-connector-aws-base/src/test/java/org/apache/flink/connector/aws/util/AWSGeneralUtilTest.java
```diff
@@ -952,4 +958,41 @@ void testNewInstanceOfDefaultCredentialsProviderCreatedForEachClient() {
...
+      .profileFile(profileFile)
+      .build();
+
+    SsoProfileCredentialsProviderFactory factory = new SsoProfileCredentialsProviderFactory();
+
+    assertThatThrownBy(() -> factory.create(context))
+      .isInstanceOf(java.io.UncheckedIOException.class)
+      .hasRootCauseInstanceOf(java.nio.file.NoSuchFileException.class)
+      .hasMessageContaining(".aws/sso/cache/");
+  }
```
leekeiabstraction, line 996: Can you elaborate why we're asserting that an exception is thrown here? What would the full exception message look like?

PR #201, opened 2025-04-16
[FLINK-35989][Connectors/AWS] Log errors on partially failed requests for AWS Kinesis Stream sink

aws-nageshvh:
## Purpose of the change

[FLINK-35989][Connectors/AWS] Log errors on partially failed requests for AWS Kinesis Stream sink

## Verifying this change

- *Added unit tests*

## Significant changes

### DIFF
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/sink/KinesisClientProvider.java
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/sink/KinesisClientProvider.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/sink/KinesisClientProvider.java
```diff
@@ -0,0 +1,39 @@
...
+interface KinesisClientProvider {
+    /**
+     * Returns a KinesisAsyncClient instance.
+     *
+     * @return The KinesisAsyncClient instance
+     */
+    KinesisAsyncClient get();
+
+    /** Closes any resources held by this provider. */
+    void close();
```
leekeiabstraction, line 38: Should the interface extend Closeable? Doing so allows for try-with-resources.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/sink/KinesisStreamsSinkWriter.java
```diff
@@ -148,6 +153,38 @@ class KinesisStreamsSinkWriter<InputT> extends AsyncSinkWriter<InputT, PutRecord
...
+                maxBufferedRequests,
+                maxBatchSizeInBytes,
+                maxTimeInBufferMS,
+                maxRecordSizeInBytes,
+                failOnError,
+                streamName,
+                streamArn,
+                kinesisClientProperties,
+                states,
+                null);
```
leekeiabstraction, line 170: Should we have an anonymous function implementation of client provider here which calls `buildClient()` instead of providing null and handling null in the other constructor method? This will allow for cleaner code in the other constructor by removing the null handling.

PR #203, opened 2025-04-26
[FLINK-37733] Externalise DynamoDB connector IT Test to E2E test package

darenwkt:
## Purpose of the change

[FLINK-37733] Externalise DynamoDB connector IT Test to E2E test package
- DynamoDB connector IT Test currently resides in `flink-connector-dynamodb` package which means:
  - It is inconsistent with other connector package where IT test resides in their own e2e test package.
  - Building the package takes longer as IT test takes time to spin up a docker environment for the test.
  - Github action/workflow does not run the IT test as they are not in e2e test package.
- This is a minor improvement to relocate the IT test into its own e2e test package similar to other connectors

## Verifying this change

- This change is a trivial rework / code cleanup without any test coverage as it just relocates existing IT test to a new E2E test package
- Verified the e2e test passes as follows:
  - Run `mvn clean verify -Prun-end-to-end-tests -DdistDir=/Users/Downloads/flink-1.19.2`
```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.apache.flink.connector.dynamodb.sink.test.DynamoDbSinkITCase
[INFO] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 114.11 s - in org.apache.flink.connector.dynamodb.sink.test.DynamoDbSinkITCase
[INFO] Running org.apache.flink.connector.dynamodb.table.test.DynamoDbDynamicSinkITCase
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 52.551 s - in org.apache.flink.connector.dynamodb.table.test.DynamoDbDynamicSinkITCase
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  02:49 min
```

## Significant changes


flink-connector-aws/flink-connector-dynamodb/src/test/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverterTest.java
```diff
@@ -216,7 +216,7 @@ void testFloat() {
         Map<String, AttributeValue> actualResult =
                 rowDataToAttributeValueConverter.convertRowData(createElement(value));
         Map<String, AttributeValue> expectedResult =
-                singletonMap(key, AttributeValue.builder().n("1.23456791E17").build());
+                singletonMap(key, AttributeValue.builder().n("1.2345679E17").build());
```
darenwkt, line 219: Updating this test case as it failed locally when I ran it with Java version 8.0.432-amzn (Corretto) due to Java floating point number precision only accurate up to 6-7 digits decimal place.

gguptp: we use java11 right?

darenwkt: > we use java11 right?

We use Java 8 based on https://repo1.maven.org/maven2/org/apache/flink/flink-connector-parent/1.0.0/flink-connector-parent-1.0.0.pom

darenwkt: Thanks @gguptp I have revisited this and found that I was running the test with java21 locally, once I changed it to java8, the test passed. As such I have reverted changes to this test case. Please take another look

gguptp (approved):

flink-connector-aws-e2e-tests/flink-connector-dynamodb-e2e-tests/pom.xml
```diff
@@ -0,0 +1,123 @@
...
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jsr310</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3</artifactId>
+            <scope>test</scope>
+        </dependency>
```
leekeiabstraction, line 109: Can you elaborate why s3 dependency is needed here?

dannycranmer:
@darenwkt can you clarify your understanding of the expectation of each module? As far as I understand the e2e tests are meant to hit real services (and require creds), while the integration tests that hit mocked services (TestContainers) reside in the connector repos (example https://github.com/apache/flink-connector-aws/blob/main/flink-connector-aws/flink-connector-kinesis/src/test/java/org/apache/flink/streaming/connectors/kinesis/FlinkKinesisITCase.java). It seems like the DDB tests are actually IT tests and based on my understanding should stay where they are. Also I see there are other IT tests in the e2e module.

@hlteoh37 what are your expectations on this?

darenwkt:
Thanks @dannycranmer for taking a look, my understanding is that we support two types of E2E tests [1]: 
 1. Localstack running on Docker
 2. Actual AWS services (and require creds).

So, the main motivation of moving these tests to E2E package type 1 for me is:
 1. It takes longer time to run these tests as part of packaging step, hence it would be more efficient to move them to E2E which runs as part of github action instead.
 2. Cleaner separation, which means the connector module only has unit tests and all IT and E2E are moved to E2E module.

However, I can understand where you are coming from as other connectors like Kafka [2] keeps their IT case in the connector module. Hence, I have no objection if the preference is to keep IT case in connector module, appreciate both you and Hong's advice on the way forward here, thank you

[1] https://github.com/apache/flink-connector-aws/blob/main/.github/workflows/common.yml#L109-L134
[2] https://github.com/apache/flink-connector-kafka/blob/main/flink-connector-kafka/src/test/java/org/apache/flink/connector/kafka/sink/ExactlyOnceKafkaWriterITCase.java

PR #206, opened 2025-05-07
[FLINK-29549]- Flink Glue Catalog integration

fmorillo7694:
## Purpose of the change

*For example: Implements the Table API for the Kinesis Source.*

## Verifying this change

*(or)*

*(or)*

*(example:)*
- *Added unit tests*

## Significant changes


flink-catalog-aws/flink-catalog-aws-glue/.gitignore
```diff
@@ -0,0 +1,2 @@
+./target
```
Samrat002, line 1: is this gitignore file required

flink-catalog-aws/flink-catalog-aws-glue/.idea/.gitignore
```diff
@@ -0,0 +1,10 @@
+# Default ignored files
```
Samrat002, line 1: Is this file required ?

flink-catalog-aws/flink-catalog-aws-glue/.idea/aws.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Samrat002, line 1: is this file required ?

flink-catalog-aws/flink-catalog-aws-glue/.idea/checkstyle-idea.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Samrat002, line 1: why do you need this file ?

Samrat002 (changes requested):
Thanks for the contribution. A few points to note:


It looks like most of the code has already been reviewed across multiple PRs.

Concerns:

1. The current implementation exposes low-level Glue details (e.g., support for lowercase table names) directly through the Flink Catalog. Is this acceptable? I recommend raising this topic with the community. In my opinion, it would be better to encapsulate such Glue-specific behavior and avoid exposing it directly via the Flink Catalog interface. The catalog's behavior should remain consistent with other catalogs, with differences controlled via configuration only.

2. Please consider implementing the basic configuration options defined in [FLIP-277](https://cwiki.apache.org/confluence/display/FLINK/FLIP-277). If that’s not feasible in this PR, a fast follow-up would be valuable, especially for users relying on different credential modes.

3. It seems that some `.idea` folder files have been committed. Please remove them from the PR.

Cheers,
Samrat

FranMorilloAWS:
> Thanks for the contribution. A few points to note:
> 
> It looks like most of the code has already been reviewed across multiple PRs.
> 
> Concerns:
> 
>     1. The current implementation exposes low-level Glue details (e.g., support for lowercase table names) directly through the Flink Catalog. Is this acceptable? I recommend raising this topic with the community. In my opinion, it would be better to encapsulate such Glue-specific behavior and avoid exposing it directly via the Flink Catalog interface. The catalog's behavior should remain consistent with other catalogs, with differences controlled via configuration only.
> 
>     2. Please consider implementing the basic configuration options defined in [FLIP-277](https://cwiki.apache.org/confluence/display/FLINK/FLIP-277). If that’s not feasible in this PR, a fast follow-up would be valuable, especially for users relying on different credential modes.
> 
>     3. It seems that some `.idea` folder files have been committed. Please remove them from the PR.
> 
> 
> Cheers, Samrat


Hey Samrat. How could we encapsulate this specific issues?  From user perspective we are already limiting them creating tables and databases with uppercase. In regards of the schema, we encapsulate storing the actual original columnNames in column parameters in Glue, so even though user sees their schema in glue with lower case (default for glue). we actually leverage the original column name.

2. Yes right after we get the first release delivered we will add the remainder authentication, and partitions support and alter support

3. Fixed!

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/GlueCatalog.java
```diff
@@ -0,0 +1,935 @@
...
+        ObjectPath normalizedPath = normalize(functionPath);
+
+        if (!databaseExists(normalizedPath.getDatabaseName())) {
+            throw new CatalogException(getName());
+        }
+
+        try {
+            return glueFunctionsOperations.getGlueFunction(normalizedPath);
+        } catch (FunctionNotExistException e) {
+            throw e;
```
leekeiabstraction, line 556: The catch and rethrow as-is is redundant. 

Do you mean to add additional information here by wrapping with another exception? Otherwise, remove the catch entirely.

leekeiabstraction (changes requested):

nicusX:
@Samrat002 @FranMorilloAWS about case sensitivity, these are the logical rules I suggest:

1. the actual case should preserved for **all supported objects** and used consistently in Flink interface, e.g. in`CREATE TABLE...` but also `SHOW TABLE`, `SHOW COLUMNS ...` etc. From a user's perspective, the way object names are actually stored in Glue should not be a concern.
2. The additional metadata should be stored in Glue in a consistent way for all object types. `Parameters` seems to be available for Database, Tables, and Columns in Glue. A problem arises for Functions which don't have any `Parameters` if I am not mistaken. 
3. The additional metadata should not leak into Flink user interface. The "originalName" should not pop up in a `SHOW TABLES` or `SHOW COLUMNS`
4. The fact Glue UI is limited and only shows lowercase names is not a concern of this component's

I would add that the way the additional metadata (the original name and any additional info required) is stored in Glue should be clearly documented, in case a user want's to build their own external interface to extract or update information in the Glue Catalog.

FranMorilloAWS:
> Thanks for the contribution. A few points to note:
> 
> It looks like most of the code has already been reviewed across multiple PRs.
> 
> Concerns:
> 
>     1. The current implementation exposes low-level Glue details (e.g., support for lowercase table names) directly through the Flink Catalog. Is this acceptable? I recommend raising this topic with the community. In my opinion, it would be better to encapsulate such Glue-specific behavior and avoid exposing it directly via the Flink Catalog interface. The catalog's behavior should remain consistent with other catalogs, with differences controlled via configuration only.
> 
>     2. Please consider implementing the basic configuration options defined in [FLIP-277](https://cwiki.apache.org/confluence/display/FLINK/FLIP-277). If that’s not feasible in this PR, a fast follow-up would be valuable, especially for users relying on different credential modes.
> 
>     3. It seems that some `.idea` folder files have been committed. Please remove them from the PR.
> 
> 
> Cheers, Samrat

Hello @Samrat002 @nicusX. Using Database Parameters and Table Parameters we are now able to use lower/upper case for defining the Database and Table Name.  By using the Show Tables/Show Databases, Describe table commands, we will show the original Flink Definition, even though in Glue UI it will be all in lower case

Samrat002:
Thanks for incorporating the changes. I will review the pr in next couple of days .

Samrat002 (commented):
Patch looks great. Thank you for making the change. 
minor comments 
Cheers, 
Samrat

fmorillo7694:
## Real AWS Glue integration test results (design rationale for `originalName`)

The latest head (b1a05d3) includes fixes driven by an integration test run against a **real Glue Data Catalog** (us-west-2), not Moto/Localstack. Posting the evidence here since it settles a design question a reviewer might reasonably raise: *why store lowercase names plus an `originalName` parameter instead of writing mixed-case names directly?*

### Empirical facts about real Glue (reproducible via the AWS CLI)

1. **Glue lowercases column names on `CreateTable`.** Creating a table with columns `userId` / `EventTime` stores and returns `userid` / `eventtime`. Glue does *not* preserve mixed-case column names, so writing them as-is silently loses case.
2. **Glue rejects column-level `Parameters` on partition columns** with `400: Parameters not supported for partition columns`. The `originalName` column-parameter mechanism therefore cannot be used for partition keys — their declared case is preserved in an order-preserving *table-level* parameter (`flink.original-partition-keys`) instead.

### What case loss breaks (observed before the fix, same suite)

- `DESCRIBE` returned `[userid, eventtime, username]` for a table declared `[userId, EventTime, userName]`
- `SELECT userId FROM t` → *Column 'userId' not found in any table; did you mean 'userid'?*
- Partition ops failed to match: spec `{eventDate=...}` vs stored keys `[eventdate]`
- Streaming `INSERT INTO` / `SELECT` against catalog-registered Kinesis tables failed to plan, and JSON field names on the wire would be lowercased (silent data corruption for consumers)

### Validation on the current head

25-step integration suite against real AWS, all passing:

- **DDL**: mixed-case databases, tables, and columns round-trip exactly (create / show / describe / drop); `ALTER TABLE ... SET` (now implemented via `UpdateTable`) with a case-preservation regression check; partitioned tables with `ADD/SHOW/DROP PARTITION` on mixed-case partition keys
- **Streaming**: unbounded job writing to and reading from a real Kinesis stream through source/sink tables registered in this catalog; JSON field names verified case-intact on the wire via `GetRecords`; `PARTITIONED BY` on a Kinesis table verified to set the record partition key from the (case-restored) column
- Unit tests: 126/126

One known behavior worth documenting: partition columns are appended after data columns on read-back (Glue models partition keys outside the storage descriptor, like Hive), so declaring partition columns last keeps the schema order stable across the round-trip.

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/GlueCatalog.java
```diff
@@ -0,0 +1,935 @@
...
+    }
+
+    /**
+     * Closes the GlueCatalog and releases resources.
+     *
+     * @throws CatalogException if an error occurs during the closing process
+     */
+    @Override
+    public void close() throws CatalogException {
+        if (glueClient != null) {
```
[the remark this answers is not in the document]
fmorillo7694: Yes - the client is only guaranteed to exist once the catalog has actually been used: a catalog that is registered and closed without ever running a DDL/DML statement can reach close() with no client, as Samrat described. The null guard stays for that case. close() itself has been simplified in 002c0fb (see the retry thread).

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/GlueCatalog.java
```diff
@@ -0,0 +1,935 @@
...
+            LOG.info("Closing GlueCatalog client");
+            int maxRetries = 3;
+            int retryCount = 0;
+            long retryDelayMs = 200;
+            while (retryCount < maxRetries) {
```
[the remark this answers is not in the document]
fmorillo7694: Good catch - you're right, the SDK close() is best-effort and doesn't surface exceptions. Removed the retry logic in 002c0fb; close() is now a plain guarded glueClient.close().

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/GlueCatalog.java
```diff
@@ -0,0 +1,935 @@
...
+            throw new DatabaseNotExistException(getName(), databaseName);
+        }
+
+        try {
+            // Get all tables in the database
+            List<Table> allTables = glueClient.getTables(builder -> builder.databaseName(databaseName))
+                    .tableList();
+
+            // Filter tables to only include those that are of type VIEW
+            List<String> viewNames = allTables.stream()
```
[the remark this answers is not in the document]
fmorillo7694: This code path was restructured in later commits and the early-return suggestion no longer maps to the current implementation (thread is marked outdated). Resolving.

docs/content/docs/connectors/table/glue.md
```diff
@@ -0,0 +1,433 @@
...
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# AWS Glue Catalog
+
+The AWS Glue Catalog provides a way to use [AWS Glue](https://aws.amazon.com/glue) as a catalog for Apache Flink. 
+This allows users to access Glue's metadata store directly from Flink SQL and Table API.
```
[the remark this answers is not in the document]
fmorillo7694: Correct - the catalog is registered through the Table API / SQL surface, and DataStream programs can use it by converting between DataStreams and Tables via StreamTableEnvironment. Added this clarification to the docs in 002c0fb.

docs/content/docs/connectors/table/glue.md
```diff
@@ -0,0 +1,433 @@
...
+## Dependencies
+
+{{< sql_download_table "glue" >}}
+
+## Prerequisites
+
+Before getting started, ensure you have the following:
+
+- **AWS account** with appropriate permissions for AWS Glue and other required services
+- **AWS credentials** properly configured
```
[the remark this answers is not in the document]
fmorillo7694: Added a link to the AWS SDK credentials configuration guide in the Prerequisites section in 002c0fb.

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/factory/GlueCatalogFactory.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public Catalog createCatalog(Context context) {
+        Map<String, String> config = context.getOptions();
+        String name = context.getName();
+        String region = config.get(REGION.key());
+        String defaultDatabase = config.getOrDefault(DEFAULT_DATABASE.key(), DEFAULT_DATABASE.defaultValue());
+
+        // Ensure required properties are present
```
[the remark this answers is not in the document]
fmorillo7694: Removed in 002c0fb.

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/operator/GlueFunctionOperator.java
```diff
@@ -0,0 +1,295 @@
...
+            throw new CatalogException(
+                String.format("Error getting function %s: %s", functionPath.getFullName(), e.getMessage()), e);
+        }
+    }
+
+    public List<String> listGlueFunctions(String databaseName) {
+        GetUserDefinedFunctionsRequest.Builder functionsRequest =
+                GetUserDefinedFunctionsRequest.builder()
+                        .databaseName(databaseName);
+        List<String> glueFunctions;
```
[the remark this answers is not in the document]
fmorillo7694: Applied in 002c0fb - the list is now initialized at declaration so it can never be null.

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/operator/GlueOperator.java
```diff
@@ -0,0 +1,45 @@
...
+    /** The catalog name associated with the Glue operations. */
+    protected final String catalogName;
+
+    /**
+     * Constructor to initialize the shared fields.
+     *
+     * @param glueClient The Glue client used for interacting with the AWS Glue service.
+     * @param catalogName The catalog name associated with the Glue operations.
+     */
+    protected GlueOperator(GlueClient glueClient, String catalogName) {
```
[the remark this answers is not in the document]
fmorillo7694: Added Preconditions.checkNotNull on both glueClient and catalogName in the GlueOperator constructor in 002c0fb.

fmorillo7694:
@Samrat002 All open review threads are now addressed and resolved as of 002c0fb:

- Simplified `GlueCatalog#close()` (removed the retry logic, per @leekeiabstraction's finding that the SDK `close()` is best-effort)
- Removed the leftover comment in `GlueCatalogFactory`
- Initialized `glueFunctions` in `GlueFunctionOperator#listGlueFunctions` so it can never be null
- Added null checks on the `GlueOperator` constructor params
- Docs: clarified DataStream API usage via `StreamTableEnvironment` and linked the AWS credentials configuration guide

The branch is also synced with `main` (0 commits behind, including the Flink 2.1.2 / connector-parent 2.0.0 bump), module poms bumped to 6.1-SNAPSHOT, missing ASF license headers added, and the new spotless formatting applied. `mvn verify` passes locally (126/126 tests, spotless + RAT clean).

The CI workflow run on the new head is waiting for approval (`action_required`) — could you approve it and take another review pass when you get a chance? Thank you!

fmorillo7694:
## Test evidence: full suite executed against real AWS Glue

Following up on the review discussions about test coverage, the three commits just pushed (`38fd97b`, `d600182`, `eae545c`) extend the test pyramid and make it executable against the real service. Evidence below is from a single `mvn verify` invocation against **real AWS Glue in eu-central-1** (screenshot to follow in the next comment).

### What was run

```
IT_CASE_GLUE_CATALOG_USE_DEFAULT_CREDENTIALS=true IT_CASE_GLUE_CATALOG_REGION=eu-central-1 \
  mvn -pl flink-catalog-aws/flink-catalog-aws-glue,flink-connector-aws-e2e-tests/flink-catalog-aws-glue-e2e-tests \
  -am verify -Prun-aws-end-to-end-tests -Dsurefire.rerunFailingTestsCount=3
```

```
[WARNING] Tests run: 127, Failures: 0, Errors: 0, Skipped: 18, Flakes: 5   <- unit suite vs REAL Glue
[INFO] Tests run: 5, ... -- in ...glue.GlueCatalogMotoITCase               <- wire-level (moto, CI tier)
[INFO] Tests run: 3, ... -- in ...glue.GlueCatalogSqlMotoITCase            <- SQL path (moto, CI tier)
[INFO] Tests run: 2, ... -- in ...glue.test.GlueCatalogEndToEndITCase      <- e2e vs REAL Glue
[INFO] BUILD SUCCESS
```

- The **18 skips** are fault-injection tests (simulated Glue server errors) — impossible against the real service by definition; they skip with that reason and run normally in CI's fake mode.
- The **5 flakes** are Glue eventual-consistency transients (`CreateDatabase` → immediate read can briefly miss; `DeleteDatabase` is asynchronous); they passed on automatic rerun and are reported by surefire rather than hidden.
- CI behaviour is **unchanged**: without credentials the unit suite runs against the in-memory fake exactly as before (verified: 127/127, zero skips), and the e2e module is compile-only.

### What the new tiers add

| Tier | Tests | Runs in CI? | Backend |
|---|---|---|---|
| Unit suite (existing, now backend-switchable) | 127 | yes (fake) | fake **or real Glue** via `GlueTestClientFactory` |
| `GlueCatalogMotoITCase` — catalog API over the real SDK wire path | 5 | yes | moto (Testcontainers) |
| `GlueCatalogSqlMotoITCase` — `CREATE CATALOG` → SPI factory → planner → wire | 3 | yes | moto (Testcontainers) |
| `GlueCatalogEndToEndITCase` — full user path incl. a data round-trip (real Flink job between catalog-registered tables) | 2 | compile-only (`-Prun-aws-end-to-end-tests` + credentials) | **real AWS Glue** |

The e2e run also covers the two behaviours only real Glue exhibits: column-name lowercasing with case restoration through the catalog, and `dropDatabase` (whose emptiness check requires the UDF API that emulators do not implement).

### Bug found and fixed by the new coverage

The e2e data round-trip immediately surfaced a contract violation no DDL-only test could reach: `GlueCatalog#getFunction` threw a bare `CatalogException` when the database did not exist, but the planner probes `getFunction` on the current database for **every SQL expression** and only falls back to built-in functions on `FunctionNotExistException` — so any expression query failed SQL validation whenever the session's current database was absent from Glue. Fixed in `38fd97b` with a regression test.

fmorillo7694:
85591f5156ecc7bf-real-glue-test-evidence

flink-catalog-aws/flink-catalog-aws-glue/.idea/jarRepositories.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Samrat002: why .idea folder exists ?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/constants/AWSGlueConfigConstants.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.flink.annotation.PublicEvolving;
+
+/** Configuration keys for AWS Glue Data Catalog service usage. */
+@PublicEvolving
+public class AWSGlueConfigConstants {
+
+    /**
+     * Configure an alternative endpoint of the Glue service for GlueCatalog to access.
+     *
+     * <p>This could be used to use GlueCatalog with any glue-compatible metastore service that has
```
Samrat002: too much java doc . keep it simple and to the point

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/factory/GlueCatalogFactory.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Override
+    public String factoryIdentifier() {
+        return "glue";
+    }
+
+    @Override
+    public Set<ConfigOption<?>> requiredOptions() {
+        Set<ConfigOption<?>> options = new HashSet<>();
+        options.add(REGION);
```
Samrat002: only region is required field ?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/factory/GlueCatalogFactory.java
```diff
@@ -0,0 +1,82 @@
...
+        Map<String, String> config = context.getOptions();
+        String name = context.getName();
+        String region = config.get(REGION.key());
+        String defaultDatabase =
+                config.getOrDefault(DEFAULT_DATABASE.key(), DEFAULT_DATABASE.defaultValue());
+
+        if (region == null || region.isEmpty()) {
+            throw new CatalogException(
+                    "The 'region' property must be specified for the Glue catalog.");
+        }
```
Samrat002: region is marked as a required field. This condition becomes redundant

fmorillo7694:
@Samrat002

fmorillo7694:
@Samrat002 Good catches, both fixed in 649f4d9: removed the tracked `.idea/` directory from the catalog module (the root `.gitignore`'s `*/.idea/` pattern doesn't match nested modules, which is how it slipped in) and restored the root `.idea/vcs.xml` this branch had accidentally deleted. Also removed the module's `META-INF/NOTICE` — the module doesn't bundle any dependencies (its bundled-deps section was empty), so per repo convention only the bundling modules carry a hand-written NOTICE and the jar NOTICE comes from maven-remote-resources.

docs/content/docs/connectors/table/glue.md
```diff
@@ -0,0 +1,440 @@
...
+SELECT event_data.EventType, event_data.eventTimestamp FROM nested_json_test;
+
+-- Accessing map fields
+SELECT metadata['source_system'] FROM nested_json_test;
+```
+
+## Limitations and Considerations
+
+1. **Case Sensitivity**: As detailed above, always use the original column names from your schema definition when querying.
+2. **AWS Service Limits**: Be aware of AWS Glue service limits that may affect your application.
```
gguptp: i am wondering if we should list the limits relevant for us?

docs/content/docs/connectors/table/glue.md
```diff
@@ -0,0 +1,440 @@
+---
+title: "AWS Glue Catalog"
+weight: 11
+type: docs
+aliases:
+  - /dev/table/connectors/glue.html
```
gguptp: please have this in content.zh as well for chinese docs

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/GlueCatalog.java
```diff
@@ -0,0 +1,1818 @@
...
+    }
+
+    /**
+     * Finds the Glue storage name for a given original database name.
+     *
+     * @param originalDatabaseName The original database name
+     * @return The Glue storage name if found, null if not found
+     * @throws CatalogException if there's an error searching
+     */
+    private String findGlueDatabaseName(String originalDatabaseName) throws CatalogException {
```
gguptp: can we rename this variable something like flinkDatabaseName? or simply databaseName?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/operator/GlueDatabaseOperator.java
```diff
@@ -0,0 +1,382 @@
...
+
+    /**
+     * Retrieves the specified database from the Glue catalog.
+     *
+     * @param originalDatabaseName The original name of the database to fetch.
+     * @return The CatalogDatabase object representing the Glue database.
+     * @throws DatabaseNotExistException If the database does not exist in the Glue catalog.
+     * @throws CatalogException If there is any error retrieving the database.
+     */
+    public CatalogDatabase getDatabase(String originalDatabaseName)
```
gguptp: currently theres only one caller of this function `getDatabase` in GlueCatalog.java and that already calls glue's `getDatabase` API to find the database name. Is there scope to optimize on the glue API calls? We might be able to save one api call here

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/GlueCatalog.java
```diff
@@ -0,0 +1,1818 @@
...
+                    LOG.warn(
+                            "Error verifying table original name for: {}.{}",
+                            glueDatabaseName,
+                            glueTableName,
+                            e);
+                }
+            }
+
+            // If direct match failed, use the existing complex search method
+            return glueTableOperations.findGlueTableName(glueDatabaseName, originalTableName);
```
gguptp: i'm a bit confused with this logic. It seems its doing the same work as the if condition. Am i missing something here?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/operator/GlueTableOperator.java
```diff
@@ -0,0 +1,554 @@
...
+    /**
+     * Checks whether a table exists in the Glue catalog by Glue storage names.
+     *
+     * @param glueDatabaseName The Glue storage name of the database where the table should exist.
+     * @param glueTableName The Glue storage name of the table to check.
+     * @return true if the table exists, false otherwise.
+     */
+    public boolean glueTableExists(String glueDatabaseName, String glueTableName) {
+        try {
+            glueClient.getTable(
```
gguptp: does this always return a table when table exists?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/operator/GlueTableOperator.java
```diff
@@ -0,0 +1,554 @@
...
+                            .databaseName(databaseName)
+                            .tableInput(tableInput)
+                            .build();
+            CreateTableResponse response = glueClient.createTable(request);
+            if (response == null
+                    || (response.sdkHttpResponse() != null
+                            && !response.sdkHttpResponse().isSuccessful())) {
+                throw new CatalogException(
+                        "Error creating table: " + databaseName + "." + tableInput.name());
+            }
```
gguptp: if glue returns an error in the response, should we catch it directly as SDK would be returning an error. How about something like:
```
try {
    // Synchronous call; throws an exception immediately if it fails
    glueClient.createTable(request); 
} catch (AlreadyExistsException e) {
    throw new CatalogException("Table " + databaseName + "." + tableInput.name() + " already exists.", e);
} catch (InvalidInputException e) {
    throw new CatalogException("Invalid schema or input parameters for table: " + tableInput.name(), e);
} catch (GlueException e) {
    // Catches any other AWS Glue service errors (e.g., 4xx or 5xx HTTP responses)
    throw new CatalogException("AWS Glue service error creating table: " + tableInput.name(), e);
} catch (SdkException e) {
    // Catches network issues, timeouts, or client-side configuration errors
    throw new CatalogException("Failed to connect or communicate with AWS Glue", e);
}
```

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/operator/GlueTableOperator.java
```diff
@@ -0,0 +1,554 @@
...
+                    UpdateTableRequest.builder()
+                            .databaseName(databaseName)
+                            .tableInput(tableInput)
+                            .build();
+            UpdateTableResponse response = glueClient.updateTable(request);
+            if (response == null
+                    || (response.sdkHttpResponse() != null
+                            && !response.sdkHttpResponse().isSuccessful())) {
+                throw new CatalogException(
+                        "Error updating table: " + databaseName + "." + tableInput.name());
```
gguptp: same here, should we rely on SDK errors instead of relying on response headers?

gguptp (commented):
Thanks @fmorillo7694 for the contribution! Overall looks solid! I have left some comments

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/GlueCatalog.java
```diff
@@ -0,0 +1,1818 @@
...
+    /**
+     * Lists all views in a specified database.
+     *
+     * @param databaseName the name of the database
+     * @return a list of view names in the database
+     * @throws DatabaseNotExistException if the database does not exist
+     * @throws CatalogException if an error occurs while listing the views
+     */
+    @Override
+    public List<String> listViews(String databaseName)
```
gguptp: we dont seem to be paginating in listViews. Is that intentional?

docs/content/docs/connectors/table/glue.md
```diff
@@ -0,0 +1,440 @@
...
+
+## Limitations and Considerations
+
+1. **Case Sensitivity**: As detailed above, always use the original column names from your schema definition when querying.
+2. **AWS Service Limits**: Be aware of AWS Glue service limits that may affect your application.
+3. **Authentication**: Ensure proper AWS credentials with appropriate permissions are available.
+4. **Region Selection**: The Glue catalog must be registered with the correct AWS region where your Glue resources exist.
+5. **Unsupported Operations**: The following operations are not currently supported:
+   - ALTER DATABASE (modifying database properties)
+   - ALTER TABLE (modifying table properties or schema)
```
gguptp: we do support alter table and add / drop partition right?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/operator/GlueTableOperator.java
```diff
@@ -0,0 +1,554 @@
...
+                    LOG.warn(
+                            "Error verifying table original name for: {}.{}",
+                            glueDatabaseName,
+                            glueTableName,
+                            e);
+                }
+            }
+
+            // If direct match failed, search all tables for original name match
+            List<String> allTables = listTables(glueDatabaseName);
```
gguptp: is there a way we can optimize the getTable API call here? The GetTables API already returns all the table details

flink-connector-aws-base/src/main/java/org/apache/flink/connector/aws/table/util/HttpClientOptionUtils.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+
+/** Class for handling AWS HTTP Client config options. */
+@PublicEvolving
+public class HttpClientOptionUtils implements TableOptionsUtils, ConfigurationValidator {
```
gguptp: whats the use of this class? Currently its only getting used in tests

flink-catalog-aws/pom.xml
```diff
@@ -0,0 +1,144 @@
...
+                <groupId>software.amazon.awssdk</groupId>
+                <artifactId>auth</artifactId>
+                <version>${aws.version}</version>
+            </dependency>
+
+            <!-- Other common dependencies -->
+            <dependency>
+                <groupId>org.json</groupId>
+                <artifactId>json</artifactId>
+                <version>20231013</version>
```
gguptp: is this getting used?

flink-catalog-aws/pom.xml
```diff
@@ -0,0 +1,144 @@
...
+            <!-- AWS dependencies -->
+            <dependency>
+                <groupId>software.amazon.awssdk</groupId>
+                <artifactId>glue</artifactId>
+                <version>${aws.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>software.amazon.awssdk</groupId>
+                <artifactId>sts</artifactId>
+                <version>${aws.version}</version>
```
gguptp: i didnt seem to find sts being used. Is this expected dependency?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/GlueCatalog.java
```diff
@@ -0,0 +1,1818 @@
...
+                }
+
+                return CatalogView.of(
+                        schemaInfo,
+                        glueTable.description(),
+                        originalQuery,
+                        expandedQuery,
+                        properties);
+            } else {
+                throw new CatalogException(
```
gguptp: it means we only support TABLE and VIEW type tables. From read of glue docs, it seems theres EXTERNAL_TABLE, EXTERNAL_VIEW, VIRTUAL_VIEW etc, if we dont want to support them, should we add that in the docs?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/GlueCatalog.java
```diff
@@ -0,0 +1,1818 @@
...
+                            return glueName;
+                        }
+                    }
+                } catch (Exception e) {
+                    LOG.warn("Error verifying database original name for: {}", glueName, e);
+                }
+            }
+
+            // If direct match failed, search all databases for original name match
+            List<String> allDatabases = glueDatabaseOperations.listDatabases();
```
gguptp: can we see if we can avoid the extra GetTable API calls and just do this with GetTables API only?

flink-catalog-aws/pom.xml
```diff
@@ -0,0 +1,144 @@
...
+        <version>6.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>flink-catalog-aws-parent</artifactId>
+    <name>Flink : Catalog : AWS : Parent</name>
+    <packaging>pom</packaging>
+
+    <properties>
+        <aws.version>2.40.3</aws.version>
+        <jackson.version>2.14.2</jackson.version>
```
gguptp: lets use these from the parent pom?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/operator/GlueFunctionOperator.java
```diff
@@ -0,0 +1,315 @@
...
+     * @param functionPath path of the function
+     * @param function Flink function to be created
+     * @throws CatalogException in case of any runtime exception
+     */
+    public void createGlueFunction(ObjectPath functionPath, CatalogFunction function)
+            throws CatalogException, FunctionAlreadyExistException {
+        UserDefinedFunctionInput functionInput = createFunctionInput(functionPath, function);
+        CreateUserDefinedFunctionRequest.Builder createUDFRequest =
+                CreateUserDefinedFunctionRequest.builder()
+                        .databaseName(functionPath.getDatabaseName())
```
gguptp: will this be glue's function name or original function name? do we need some tests for this?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/GlueCatalog.java
```diff
@@ -0,0 +1,1818 @@
...
+ * GlueCatalog is an implementation of the Flink AbstractCatalog that interacts with AWS Glue. This
+ * class allows Flink to perform various catalog operations such as creating, deleting, and
+ * retrieving databases and tables from Glue. It encapsulates AWS Glue's API and provides a
+ * Flink-compatible interface.
+ *
+ * <p>This catalog uses GlueClient to interact with AWS Glue services, and operations related to
+ * databases and tables are delegated to respective helper classes like GlueDatabaseOperations and
+ * GlueTableOperations.
+ */
+public class GlueCatalog extends AbstractCatalog {
```
gguptp: @PublicEvolving?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/factory/GlueCatalogFactory.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.flink.table.catalog.exceptions.CatalogException;
+import org.apache.flink.table.catalog.glue.GlueCatalog;
+import org.apache.flink.table.factories.CatalogFactory;
+
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/** Factory for creating GlueCatalog instances. */
+public class GlueCatalogFactory implements CatalogFactory {
```
gguptp: @PublicEvolving?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/GlueCatalog.java
```diff
@@ -0,0 +1,1818 @@
...
+                CatalogTable.newBuilder()
+                        .schema(catalogView.getUnresolvedSchema())
+                        .comment(catalogView.getComment())
+                        .partitionKeys(Collections.emptyList())
+                        .options(tableProperties)
+                        .build();
+
+        // Build table input with proper name preservation
+        TableInput baseTableInput =
+                glueTableOperations.buildTableInput(
```
gguptp: here, glueColumns and storageDescriptor are same columns and will lead to duplicate columns and will lead to a validationexception. Do we need a test for this particular edge case.

How about:

```
glueTableOperations.buildTableInput(tableName, Collections.emptyList(), tempTable, storageDescriptor, tableProperties);
```

flink-catalog-aws/flink-catalog-aws-glue/src/test/java/org/apache/flink/table/catalog/glue/operator/FakeGlueClient.java
```diff
@@ -0,0 +1,511 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A mock implementation of the AWS Glue client for testing purposes. This class simulates the
+ * behavior of the real AWS Glue service without making actual API calls. It manages in-memory
+ * storage of databases, tables, and functions for testing the Glue catalog implementation.
+ */
+public class FakeGlueClient implements GlueClient {
```
gguptp: we dont seem to be testing pagination in many functions. Do we need to add support for pagination?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/factory/GlueCatalogFactory.java
```diff
@@ -0,0 +1,82 @@
...
+    @Override
+    public Set<ConfigOption<?>> optionalOptions() {
+        Set<ConfigOption<?>> options = new HashSet<>();
+        options.add(DEFAULT_DATABASE);
+        return options;
+    }
+
+    @Override
+    public Catalog createCatalog(Context context) {
+        Map<String, String> config = context.getOptions();
```
gguptp: do we need a `FactoryUtil.createCatalogFactoryHelper(this, context).validate()`? to ensure invalid options dont get ignored silently?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/GlueCatalog.java
```diff
@@ -0,0 +1,1818 @@
...
+            } else {
+                throw new CatalogException(
+                        "Unsupported table kind: " + catalogBaseTable.getTableKind());
+            }
+            LOG.info(
+                    "Successfully created {}.{} of kind {}",
+                    originalDatabaseName,
+                    originalTableName,
+                    catalogBaseTable.getTableKind());
+        } catch (Exception e) {
```
gguptp: what about `DatabaseNotExistException`?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/GlueCatalog.java
```diff
@@ -0,0 +1,1818 @@
...
+            List<String> partitionValues =
+                    toOrderedPartitionValues(tableRef, objectPath, catalogPartitionSpec);
+            gluePartitionOperations.dropPartition(
+                    tableRef.databaseName, tableRef.tableName, partitionValues);
+        } catch (software.amazon.awssdk.services.glue.model.EntityNotFoundException
+                | PartitionNotExistException
+                | TableNotExistException
+                | TableNotPartitionedException e) {
+            if (!ignoreIfNotExists) {
+                throw new PartitionNotExistException(getName(), objectPath, catalogPartitionSpec);
```
gguptp: we are catching TableNotExistException and TableNotPartitionedException and throwing PartitionNotExistException. Can this cause issues during debugging?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/constants/GlueCatalogConstants.java
```diff
@@ -0,0 +1,63 @@
...
+package org.apache.flink.table.catalog.glue.constants;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.table.catalog.glue.GlueCatalog;
+
+import java.util.regex.Pattern;
+
+/** Constants and Defined Values used for {@link GlueCatalog}. */
+@Internal
+public class GlueCatalogConstants {
```
gguptp: we have util/GlueCatalogConstants.java. Do we need this?

docs/content/docs/connectors/table/glue.md
```diff
@@ -0,0 +1,440 @@
...
+- Compatibility with Flink's Table API and SQL interface
+
+The Glue Catalog is registered through the Table API / SQL. DataStream applications can also
+use it by converting between DataStreams and Tables with the
+[DataStream API integration]({{< ref "docs/dev/table/data_stream_api" >}}), so tables backed by
+Glue metadata are accessible from DataStream programs through a `StreamTableEnvironment`.
+
+## Dependencies
+
+{{< sql_download_table "glue" >}}
```
gguptp: do we mean `sql_connector_download_table` here?

flink-catalog-aws/flink-catalog-aws-glue/src/test/java/org/apache/flink/table/catalog/glue/GlueCatalogTest.java
```diff
@@ -0,0 +1,1131 @@
...
+                .isInstanceOf(NullPointerException.class);
+
+        assertThatThrownBy(
+                        () -> {
+                            glueCatalog.createTable(new ObjectPath("db", "table"), null, false);
+                        })
+                .isInstanceOf(NullPointerException.class);
+    }
+
+    @Test
```
gguptp: lets have a round trip test as well

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/util/GlueFunctionsUtil.java
```diff
@@ -0,0 +1,96 @@
...
+        } else if (glueFunction
+                .className()
+                .startsWith(GlueCatalogConstants.FLINK_PYTHON_FUNCTION_PREFIX)) {
+            return FunctionLanguage.PYTHON;
+        } else if (glueFunction
+                .className()
+                .startsWith(GlueCatalogConstants.FLINK_SCALA_FUNCTION_PREFIX)) {
+            return FunctionLanguage.SCALA;
+        } else {
+            throw new CatalogException(
```
gguptp: are these 3 the only cases. What about spark and hive?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/util/GlueTableUtils.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /**
+     * Converts a Flink column to a Glue column. The column's data type is converted using the
+     * GlueTypeConverter.
+     *
+     * @param flinkColumn The Flink column to be converted.
+     * @return The corresponding Glue column.
+     */
+    public Column mapFlinkColumnToGlueColumn(org.apache.flink.table.catalog.Column flinkColumn) {
```
gguptp: we're only persisting column name and type, what if there are watermarks in the flink schema? they'll get lost on a read back from glue. Same with primary keys?

flink-catalog-aws/flink-catalog-aws-glue/src/main/java/org/apache/flink/table/catalog/glue/factory/GlueCatalogFactory.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.flink.table.catalog.Catalog;
+import org.apache.flink.table.catalog.exceptions.CatalogException;
+import org.apache.flink.table.catalog.glue.GlueCatalog;
+import org.apache.flink.table.factories.CatalogFactory;
+
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/** Factory for creating GlueCatalog instances. */
```
gguptp: we might also want a UT for this

PR #208, opened 2025-05-16
[FLINK-34071][Connectors/Kinesis] Decouple NettyEventLoop thread's onNext() by handing over blocking queue put to a separate executor. Using a shared executor across shards to execute processing of event received by NettyEventLoop.

aws-nageshvh:
## Purpose of the change

This PR decouples the NettyEventLoop thread from the blockingQueue.put operation in the Kinesis Enhanced Fan-Out (EFO) connector. Previously, the critical path operations (queue.put, startingPosition recording, and requestRecords) were executed directly in the NettyEventLoop thread, which could lead to blocking and potential performance issues.

In this implementation, we extract the blocking operations into a separate method that is executed by a shared executor service, allowing the NettyEventLoop thread to return quickly while the potentially blocking operations are handled asynchronously. This improves the overall performance and stability of the Kinesis EFO connector, especially under high load conditions.  This pattern is also in accordance with the reactive streams spec for implementing onNext() as provided here - https://github.com/reactive-streams/reactive-streams-jvm/blob/master/README.md#2-subscriber-code

It also solves the deadlock issue reported in FLINK-34071 which was happening because the NettyEventLoop thread on sink was blocked by NettyEventLoop thread on source. However, the NettyEventLoop on source isn't released because it's depending on sink to relieve the backpressure.

## Verifying this change
This change added unit tests and can be verified as follows:
- *Added and improved unit tests that cover various failure scenarios*
- *Manually verified by running the Kinesis connector on a production Flink cluster with a deterministic reproducible deadlock scenario under backpressure*

## Significant changes
- This change is an internal implementation improvement that doesn't affect the public API or serialization. It enhances the performance and stability of the existing Kinesis EFO connector by optimizing how the NettyEventLoop thread interacts with the blocking queue operations.

### DIFF
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
@@ -91,6 +91,27 @@ public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
      */
     private final ExecutorService sharedShardSubscriptionExecutor;
 
+    /**
+     * Shared executor service for making subscribeToShard API calls.
+     *
+     * <p>This executor is separate from the event processing executor to avoid contention
+     * between API calls and event processing. Using a dedicated executor for subscription calls
+     * provides several important benefits:
+     *
+     * <ol>
+     *   <li>Prevents blocking of the main thread or event processing threads during API calls</li>
+     *   <li>Isolates API call failures from event processing operations</li>
+     *   <li>Allows for controlled concurrency of API calls across multiple shards</li>
+     *   <li>Prevents potential deadlocks that could occur when the same thread handles both
+     *       subscription calls and event processing</li>
+     * </ol>
+     *
+     * <p>The executor uses a smaller number of threads than the event processing executor since
+     * subscription calls are less frequent and primarily I/O bound. This helps optimize resource
+     * usage while still providing sufficient parallelism for multiple concurrent subscription calls.
+     */
+    private final ExecutorService sharedSubscriptionCallExecutor;
+
     private final Map<String, FanOutKinesisShardSubscription> splitSubscriptions = new HashMap<>();
 
     /**
@@ -104,7 +125,8 @@ public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
                 String shardId,
                 StartingPosition startingPosition,
                 Duration timeout,
-                ExecutorService executor);
+                ExecutorService eventProcessingExecutor,
+                ExecutorService subscriptionCallExecutor);
     }
 
     /**
@@ -118,14 +140,16 @@ public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
                 String shardId,
                 StartingPosition startingPosition,
                 Duration timeout,
-                ExecutorService executor) {
+                ExecutorService eventProcessingExecutor,
+                ExecutorService subscriptionCallExecutor) {
             return new FanOutKinesisShardSubscription(
                     proxy,
                     consumerArn,
                     shardId,
                     startingPosition,
                     timeout,
-                    executor);
+                    eventProcessingExecutor,
+                    subscriptionCallExecutor);
         }
     }
 
@@ -152,18 +176,20 @@ public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
             shardMetricGroupMap,
             configuration,
             subscriptionFactory,
-            createDefaultExecutor());
+            createDefaultEventProcessingExecutor(),
+            createDefaultSubscriptionCallExecutor());
     }
 
     /**
-     * Constructor with injected executor service for testing.
+     * Constructor with injected executor services for testing.
      *
      * @param asyncStreamProxy The proxy for Kinesis API calls
      * @param consumerArn The ARN of the consumer
      * @param shardMetricGroupMap The metrics map
      * @param configuration The configuration
      * @param subscriptionFactory The factory for creating subscriptions
-     * @param executorService The executor service to use for subscription tasks
+     * @param eventProcessingExecutor The executor service to use for event processing tasks
+     * @param subscriptionCallExecutor The executor service to use for subscription API calls
      */
     @VisibleForTesting
     FanOutKinesisShardSplitReader(
@@ -172,22 +198,24 @@ public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
             Map<String, KinesisShardMetrics> shardMetricGroupMap,
             Configuration configuration,
             SubscriptionFactory subscriptionFactory,
-            ExecutorService executorService) {
+            ExecutorService eventProcessingExecutor,
+            ExecutorService subscriptionCallExecutor) {
         super(shardMetricGroupMap, configuration);
         this.asyncStreamProxy = asyncStreamProxy;
         this.consumerArn = consumerArn;
         this.subscriptionTimeout = configuration.get(EFO_CONSUMER_SUBSCRIPTION_TIMEOUT);
         this.deregisterTimeout = configuration.get(EFO_DEREGISTER_CONSUMER_TIMEOUT);
         this.subscriptionFactory = subscriptionFactory;
-        this.sharedShardSubscriptionExecutor = executorService;
+        this.sharedShardSubscriptionExecutor = eventProcessingExecutor;
+        this.sharedSubscriptionCallExecutor = subscriptionCallExecutor;
     }
 
     /**
-     * Creates the default executor service for subscription tasks.
+     * Creates the default executor service for event processing tasks.
      *
      * @return A new executor service
      */
-    private static ExecutorService createDefaultExecutor() {
+    private static ExecutorService createDefaultEventProcessingExecutor() {
         int minThreads = Runtime.getRuntime().availableProcessors();
         int maxThreads = minThreads * 2;
         return new ThreadPoolExecutor(
@@ -196,6 +224,37 @@ public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
             60L, TimeUnit.SECONDS,
             new LinkedBlockingQueue<>(), // Unbounded queue with natural flow control
             new ExecutorThreadFactory("kinesis-efo-subscription"));
+    }
+
+    /**
+     * Creates the default executor service for subscription API calls.
+     *
+     * <p>This executor is configured with:
+     * <ul>
+     *   <li>Minimum threads: 1 - Ensures at least one thread is always available for API calls</li>
+     *   <li>Maximum threads: max(2, availableProcessors/4) - Scales with system resources but
+     *       keeps the thread count relatively low since API calls are I/O bound</li>
+     *   <li>Keep-alive time: 60 seconds - Allows for efficient reuse of threads</li>
+     *   <li>Unbounded queue - Safe because the number of subscription tasks is naturally limited
+     *       by the number of shards</li>
+     *   <li>Custom thread factory - Provides meaningful thread names for debugging</li>
+     * </ul>
+     *
+     * <p>This configuration balances resource efficiency with responsiveness for subscription calls.
+     * Since subscription calls are primarily waiting on network I/O, a relatively small number of
+     * threads can efficiently handle many concurrent calls.
+     *
+     * @return A new executor service optimized for subscription API calls
+     */
+    private static ExecutorService createDefaultSubscriptionCallExecutor() {
+        int minThreads = 1;
+        int maxThreads = Math.max(2, Runtime.getRuntime().availableProcessors() / 4);
+        return new ThreadPoolExecutor(
+            minThreads,
+            maxThreads,
+            60L, TimeUnit.SECONDS,
+            new LinkedBlockingQueue<>(), // Unbounded queue with natural flow control
+            new ExecutorThreadFactory("kinesis-subscription-caller"));
     }
 
     @Override
@@ -226,7 +285,8 @@ public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
                             split.getShardId(),
                             split.getStartingPosition(),
                             subscriptionTimeout,
-                            sharedShardSubscriptionExecutor);
+                            sharedShardSubscriptionExecutor,
+                            sharedSubscriptionCallExecutor);
             subscription.activateSubscription();
             splitSubscriptions.put(split.splitId(), subscription);
         }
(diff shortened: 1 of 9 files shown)


flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -334,4 +368,87 @@ public void onComplete() {
...
+     *
+     * @param event The subscription event to process
+     */
+    private void submitEventProcessingTask(SubscribeToShardEvent event) {
+        try {
+            subscriptionEventProcessingExecutor.execute(() -> {
+                synchronized (subscriptionEventProcessingLock) {
+                    try {
+                        processSubscriptionEvent(event);
+                    } catch (Exception e) {
```
leekeiabstraction, line 384: processSubscriptionEvent() can be long running here. Do we need to catch and handle InterruptedException?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -334,4 +368,87 @@ public void onComplete() {
...
+     * <p>These operations are executed sequentially for each shard to ensure thread safety
+     * and prevent race conditions. The bounded nature of the event queue (capacity 2) combined
+     * with only requesting more records after processing an event provides natural flow control,
+     * effectively limiting the number of tasks in the executor's queue.
+     *
+     * <p>This method is made public for testing purposes.
+     *
+     * @param event The subscription event to process
+     */
+    public void processSubscriptionEvent(SubscribeToShardEvent event) {
```
leekeiabstraction, line 418: What is the motivation for this method to be public?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -334,4 +368,87 @@ public void onComplete() {
...
+                    } catch (Exception e) {
+                        // For critical path operations, propagate exceptions to cause a Flink job restart
+                        LOG.error("Error processing subscription event", e);
+                        // Propagate the exception to the subscription exception handler
+                        terminateSubscription(new KinesisStreamsSourceException(
+                            "Error processing subscription event", e));
+                    }
+                }
+            });
+        } catch (Exception e) {
```
leekeiabstraction, line 393: Should we be more precise with the exception that we're catching, for example RejectedExecutionException?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -334,4 +368,87 @@ public void onComplete() {
...
+    /**
+     * Submits an event processing task to the executor service.
+     * This method encapsulates the task submission logic and error handling.
+     *
+     * @param event The subscription event to process
+     */
+    private void submitEventProcessingTask(SubscribeToShardEvent event) {
+        try {
+            subscriptionEventProcessingExecutor.execute(() -> {
+                synchronized (subscriptionEventProcessingLock) {
```
leekeiabstraction, line 381: The use of lock object can be avoided if the method `processSubscriptionEvent` be defined with `synchronized`.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
```diff
@@ -45,17 +52,128 @@ public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
...
+     * @return A new executor service
+     */
+    private static ExecutorService createDefaultExecutor() {
+        int minThreads = Runtime.getRuntime().availableProcessors();
+        int maxThreads = minThreads * 2;
+        return new ThreadPoolExecutor(
+            minThreads,
+            maxThreads,
+            60L, TimeUnit.SECONDS,
+            new LinkedBlockingQueue<>(), // Unbounded queue with natural flow control
```
leekeiabstraction, line 175: Can you elaborate what will happen in the case where downstream is severely backpressured?

I believe there is a risk of OutOfMemoryError. When there is backpressure downstream, all threads in thread pool will be blocked during `eventQueue.put(event)` in `processSubscriptionEvent()` as the eventQueue is filled up. The ThreadPoolExecutor's LinkedBlockingQueue will then keep growing eventually causing an OOM.

I can see a scenario where i) Flink job reprocesses from trim horizon with hours worth of data ii) The Flink job initial checkpoint is very slow or large causing backpressure for long time e.g. 30 minutes. Under this scenario, the connector may OOM repeatedly and the job will be unrecoverable. 

Do we need to propagate the backpressure further upstream by delaying new subscription when LinkedBlockingQueue grows too large?

leekeiabstraction (changes requested):

elrob:
Thanks for working on this change.
I have a suspicion that this issue or something similar is the cause of this:
https://issues.apache.org/jira/browse/FLINK-37648

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
```diff
@@ -19,43 +19,242 @@
...
+    /**
+     * Shared executor service for all shard subscriptions.
+     *
+     * <p>This executor uses an unbounded queue ({@link LinkedBlockingQueue}) but this does not pose
+     * a risk of out-of-memory errors due to the natural flow control mechanisms in the system:
+     *
+     * <ol>
+     *   <li>Each {@link FanOutKinesisShardSubscription} has a bounded event queue with capacity of 2</li>
+     *   <li>New records are only requested after processing an event (via {@code requestRecords()})</li>
+     *   <li>When a shard's queue is full, the processing thread blocks at the {@code put()} operation</li>
```
leekeiabstraction, line 74: I think I'm missing context to interpret how the backpressure works. 

Can you clarify if put() and requestRecords() is done in order and requestRecords() is done only after put()? If so, that was the information that I was missing.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
```diff
@@ -19,43 +19,242 @@
...
+    public interface SubscriptionFactory {
+        FanOutKinesisShardSubscription createSubscription(
+                AsyncStreamProxy proxy,
+                String consumerArn,
+                String shardId,
+                StartingPosition startingPosition,
+                Duration timeout,
+                ExecutorService eventProcessingExecutor,
+                ExecutorService subscriptionCallExecutor);
+    }
```
leekeiabstraction, line 130: Should SubscriptionFactory be an interface in FanOutKinesisShardSubscription or in its own file instead?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
```diff
@@ -19,43 +19,242 @@
...
+     * <p>Where:
+     * <ul>
+     *   <li>2 * Number_of_Shards: Total capacity of all shard event queues</li>
+     *   <li>min(Number_of_Shards, Number_of_Threads): Maximum events being actively processed</li>
+     * </ul>
+     *
+     * <p>This ensures that memory usage scales linearly with the number of shards, not exponentially,
+     * making it safe to use an unbounded executor queue even with a large number of shards.
+     */
+    private final ExecutorService sharedShardSubscriptionExecutor;
```
leekeiabstraction, line 92: Let's make the name of the event executor service consistent, there are currently 2 names.. `sharedShardSubscriptionExecutor` and `eventProcessingExecutor`;

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
```diff
@@ -19,43 +19,242 @@
...
+     */
+    private static ExecutorService createDefaultEventProcessingExecutor() {
+        int minThreads = Runtime.getRuntime().availableProcessors();
+        int maxThreads = minThreads * 2;
+        return new ThreadPoolExecutor(
+            minThreads,
+            maxThreads,
+            60L, TimeUnit.SECONDS,
+            new LinkedBlockingQueue<>(), // Unbounded queue with natural flow control
+            new ExecutorThreadFactory("kinesis-efo-subscription"));
```
leekeiabstraction, line 226: Let's make the threads' names consistent with the executor name e.g. event-processing-executor-service-thread

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
```diff
@@ -19,43 +19,242 @@
...
+     */
+    private static ExecutorService createDefaultSubscriptionCallExecutor() {
+        int minThreads = 1;
+        int maxThreads = Math.max(2, Runtime.getRuntime().availableProcessors() / 4);
+        return new ThreadPoolExecutor(
+            minThreads,
+            maxThreads,
+            60L, TimeUnit.SECONDS,
+            new LinkedBlockingQueue<>(), // Unbounded queue with natural flow control
+            new ExecutorThreadFactory("kinesis-subscription-caller"));
```
leekeiabstraction, line 257: Let's make the threads' names consistent with the executor name e.g. subscription-call-executor-service-thread

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
```diff
@@ -80,19 +279,167 @@ public void handleSplitsChanges(SplitsChange<KinesisShardSplit> splitsChanges) {
...
+     * The longer timeout accounts for potential network delays during shutdown.
+     */
+    private void closeAsyncStreamProxy() {
+        // Create a dedicated single-threaded executor for closing the asyncStreamProxy
+        // This prevents the close operation from being affected by the main executor shutdown
+        ExecutorService closeExecutor = new ThreadPoolExecutor(
+            1, 1,
+            0L, TimeUnit.MILLISECONDS,
+            new LinkedBlockingQueue<>(),
+            new ExecutorThreadFactory("kinesis-client-close"));
```
leekeiabstraction, line 413: Let's use more elaborate thread names e.g. close-async-stream-proxy-executor-service

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
```diff
@@ -80,19 +279,167 @@ public void handleSplitsChanges(SplitsChange<KinesisShardSplit> splitsChanges) {
...
+
+    /**
+     * Cancels all active subscriptions to prevent new events from being processed.
+     *
+     * <p>After cancelling subscriptions, we wait a short time to allow the cancellation
+     * signals to propagate before proceeding with executor shutdown.
+     */
+    private void cancelActiveSubscriptions() {
+        for (FanOutKinesisShardSubscription subscription : splitSubscriptions.values()) {
+            if (subscription.isActive()) {
```
leekeiabstraction, line 325: Is this if necessary? cancelSubscription() already check if subscription is active.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -132,27 +217,49 @@ public void activateSubscription() {
...
+                    LOG.error("Exception during subscribeToShard API call for shard {}", shardId, e);
+                    terminateSubscription(e);
+                    waitForSubscriptionLatch.countDown();
+                    return CompletableFuture.<Void>completedFuture(null);
+                }
+            },
+            subscriptionCallExecutor
+        ).thenCompose(future -> future); // Flatten the CompletableFuture<CompletableFuture<Void>> to CompletableFuture<Void>
+
+        subscriptionFuture.exceptionally(
```
leekeiabstraction, line 245: Should we retain the previous comment?

```
        // We don't need to keep track of the future here because we monitor subscription success
        // using our own CountDownLatch
```

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -262,17 +373,21 @@ private FanOutShardSubscriber(CountDownLatch subscriptionLatch) {
         }
 
         public void requestRecords() {
-            subscription.request(1);
+            if (subscription != null) {
+                subscription.request(1);
+            } else {
+                LOG.warn("Cannot request records - subscription is null for shard {}", shardId);
```
leekeiabstraction, line 379: Under what condition does this happen? If not handled correctly, wouldn't this cause a subscription to stop getting further data? Should we just let it throw NPE here instead?

PR #209, opened 2025-05-24
[FLINK-37841][Connectors/DynamoDB] Using standard retry strategy instead of adaptive retry strategy to build ddb streams client

gguptp:
## Purpose of the change

AWS docs say the following:

> The adaptive retry strategy includes all the features of the standard strategy and adds a client-side rate limiter that measures the rate of throttled requests compared to non-throttled requests. The strategy uses this measurement to slow down the requests in an attempt to stay within a safe bandwidth, ideally causing zero throttling errors.

 

This means requests maybe getting slowed down and the connector not achieving desired performance in things like shard discovery. Having standard retry strategy should be able to better fix this issue.

## Verifying this change

Tested on local flink cluster

*(or)*

*(or)*

*(example:)*
- *Added unit tests*

Manual verification of the change was done

## Significant changes


gguptp:
Hi @darenwkt while i agree using adaptive retry strategy prevents throttling better than standard, it does so by throttling at client side. This can cause api calls to delay. For example, from my testing what i noticed is that if i used adaptive strategy, the difference between two api calls comes out to be 200 millis which can be quite a lot. 

AWS recommends standard retry strategy to be good for almost all the usecases until we have something specific. We have a backlog task in jira already to allow customers to expose the retry strategy

darenwkt (approved):
Thanks Abhi, approved

gguptp:
@hlteoh37 could you help in reviewing this PR please

gguptp:
Will update this PR, needs some changes to be made

PR #215, opened 2025-06-16
FLINK-37949: Avoid netty deadlock with graceful shutdown

jayteej:
https://issues.apache.org/jira/browse/FLINK-37949 shows that a deadlock in the netty client can cause savepoint delay. This change ensures that any pending operations are terminated before closing down the client and also updates the latest minor version of netty in case this is a known deadlock issue.


leekeiabstraction (approved):

jayteej:
> LGTM @jayteej Thanks for the contribution!
> 
> Some housekeeping, if that's ok! 👀
> 
> 1. CI is failing due to formatting errors, do you mind running `mvn spotless:apply` to fix them?
> 2. The title of the commit is not correct - can we squash into 1 and follow the format of `[FLINK-37949][Connectors/Kinesis] xxxx` ?

Sorry for the delay. I didn't notice u responded. I've had to refactor this as the trunk changed the file structure quite a lot.

PR #224, opened 2025-12-17
[FLINK-38814][Connector/Kinesis] Kinesis Source in EFO - Fix idle consumer on subscription failure

nadberezny:
**Purpose of the Change**
This PR fixes a critical stability issue in the FanOutKinesisShardSubscription where the connector could enter a "zombie" state—appearing active but not consuming events—following a failed subscription attempt (ResourceInUseException).

It also resolves some potential duplicates due to non thread safe shard starting position.

More details can be found in the JIRA ticket description:
https://issues.apache.org/jira/browse/FLINK-38814

### DIFF
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
@@ -85,8 +85,9 @@ public class FanOutKinesisShardSubscription {
 
     // Store the current starting position for this subscription. Will be updated each time new
     // batch of records is consumed
-    private StartingPosition startingPosition;
-    private FanOutShardSubscriber shardSubscriber;
+    // Cross-thread access: written by Netty thread in Subscriber#onNext, read by Flink main thread in kinesis#subscribeToShard.
+    private volatile StartingPosition startingPosition;
+    private volatile FanOutShardSubscriber shardSubscriber;
 
     public FanOutKinesisShardSubscription(
             AsyncStreamProxy kinesis,
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
@@ -138,17 +138,11 @@ public class FanOutKinesisShardSubscription {
         kinesis.subscribeToShard(consumerArn, shardId, startingPosition, responseHandler)
                 .exceptionally(
                         throwable -> {
-                            // If consumer exists and is still activating, we want to countdown.
-                            if (ExceptionUtils.findThrowable(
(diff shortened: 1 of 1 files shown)


AndyShoeshim:
java.util.concurrent.TimeoutException: Timeout when subscribing to shard shardId-000000000033 with starting position StartingPosition{shardIteratorType=AFTER_SEQUENCE_NUMBER, startingMarker=49670023483232461788730612813735157811693693431491592722} 


can this exception also be related to this? I have a similar use case described in the issue, in which the consumer gets in a zombie like state

AndyShoeshim:
Any news on this? As of now, with this kind of issue the EFO consumer on AWS isn't an option in any production environment

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -85,8 +85,9 @@ public class FanOutKinesisShardSubscription {
 
     // Store the current starting position for this subscription. Will be updated each time new
     // batch of records is consumed
-    private StartingPosition startingPosition;
-    private FanOutShardSubscriber shardSubscriber;
+    // Cross-thread access: written by Netty thread in Subscriber#onNext, read by Flink main thread in kinesis#subscribeToShard.
+    private volatile StartingPosition startingPosition;
+    private volatile FanOutShardSubscriber shardSubscriber;
```
gguptp, line 90: i dont think these need to be volatile since they are getting populated by the reader

gguptp:
@nadberezny could we add some UTs for this behaviour?

Also, would be good to define in the jira on what would happen if ResourceInUseException comes and whether any reader will be able to read from the shard or not

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -161,14 +153,16 @@ public void activateSubscription() {
...
-                            shardSubscriber.requestRecords();
+                            Throwable error = subscriptionException.get();
+                            if (error == null) {
+                                LOG.info(
+                                        "Successfully subscribed to shard {} with starting position {} for consumer {}.",
+                                        shardId,
+                                        startingPosition,
+                                        consumerArn);
+                            } else {
+                                LOG.debug(String.format("Initialization finished with error: %s", error.getMessage()), error);
```
davidradl, line 164: Why is this debug - I would expect errors to be logged as errors - or at least as high as the none error case?
I notice LOG.info has {} and the debug used String.format( why are they using different methods to bring in parameters?

Curious why does it  talk about Initialization- I would have expected something like- Failed to subscribe to shard and put out the shardId and other useful identifying information as well as the error in the logging - similar to the timeout code. 

Why do we not throw an exception in the Error case - like we do for the timeout case? If this is intended I suggest putting a comment in explaining.

tim-james-expel:
Hi folks!  Is this still happening?  Is `KinesisStreamsSource` safe or unsafe to use in production with Kinesis EFO without this fix? 

> This PR fixes a critical stability issue

Sounds pretty important.

Thank you!

chenylee-aws:
This issue is addressed in this PR https://github.com/apache/flink-connector-aws/pull/247 and released in 6.0.1. This PR can be closed. Thanks.

PR #234, opened 2026-02-18
[FLINK-39106] Support ROW and ARRAY<ROW> data types in DynamoDB API / SQL sink

ddebowczyk92:
## Purpose of the change

Adds support for ROW and ARRAY<ROW> data types in the DynamoDB Table API / SQL sink. DynamoDB's Map (M) / document attribute type is a natural fit for Flink's `ROW` type, but was previously unsupported - the converter would fall through to `EnhancedType.of(RowData.class)` which has no built-in `AttributeConverter`, causing a runtime failure. Recursive nesting (`ROW` within `ROW`, `ARRAY<ROW>` within `ROW`) is fully supported.

## Verifying this change

- Added `testRowDataType` — single-level nested `ROW` with `STRING` and `INT` fields
- Added `testRowDataTypeNullInnerField` — null field inside nested `ROW` renders as `AttributeValue(NUL=true)`
- Added `testNestedRowDataType` — 3-level deep nesting (outer -> middle -> leaf)
- Added `testRowDataTypeArray` — `ARRAY<ROW<name STRING, value INT>>` with 2 elements, asserts DynamoDB List of Maps
- Added `ROW` and `ARRAY<ROW>` columns to `DynamoDbDynamicSinkFactoryTest` schema and IT case SQL definition

## Significant changes
  - If yes, how is this documented? (not documented)

### DIFF
--- a/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverter.java
+++ b/flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverter.java
@@ -40,6 +40,7 @@ import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.IntStream;
 
 import static org.apache.flink.table.data.RowData.createFieldGetter;
@@ -108,21 +109,21 @@ public class RowDataToAttributeValueConverter {
                 });
     }
 
-    private java.util.Optional<AttributeConverter> buildRowAttributeConverter(DataType dataType) {
-        if (LogicalTypeRoot.ROW.equals(dataType.getLogicalType().getTypeRoot())) {
-            return java.util.Optional.of(createRowDocumentConverter(buildRowTableSchema(dataType)));
+    private Optional<AttributeConverter> buildRowAttributeConverter(DataType dataType) {
+        if (LogicalTypeRoot.ROW == dataType.getLogicalType().getTypeRoot()) {
+            return Optional.of(createRowDocumentConverter(buildRowTableSchema(dataType)));
         }
         if (dataType instanceof CollectionDataType) {
             DataType elementDataType = ((CollectionDataType) dataType).getElementDataType();
             if (LogicalTypeRoot.ROW.equals(elementDataType.getLogicalType().getTypeRoot())) {
                 AttributeConverter<Row> elementConverter =
                         createRowDocumentConverter(buildRowTableSchema(elementDataType));
-                return java.util.Optional.of(
+                return Optional.of(
                         new ArrayAttributeConverter<>(
                                 elementConverter, EnhancedType.of(Row[].class)));
             }
         }
-        return java.util.Optional.empty();
+        return Optional.empty();
     }
 
     private static AttributeConverter<Row> createRowDocumentConverter(
@@ -170,11 +171,11 @@ public class RowDataToAttributeValueConverter {
         builder.attributeConverterProviders(
                 newAttributeConverterProvider, AttributeConverterProvider.defaultProvider());
 
-        DataTypes.Field[] fields = DataType.getFields(dataType).toArray(new DataTypes.Field[0]);
-        IntStream.range(0, fields.length)
+        final List<DataTypes.Field> fields = DataType.getFields(dataType);
+        IntStream.range(0, fields.size())
                 .forEach(
                         idx -> {
-                            final DataTypes.Field field = fields[idx];
+                            final DataTypes.Field field = fields.get(idx);
                             final DataType fieldDataType = field.getDataType();
                             builder.addAttribute(
                                     getEnhancedType(fieldDataType),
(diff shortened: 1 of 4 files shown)


flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverter.java
```diff
@@ -84,18 +91,64 @@ private StaticTableSchema.Builder<RowData> addAttribute(
...
+                        new ArrayAttributeConverter<>(
+                                elementConverter, EnhancedType.of(Row[].class)));
+            }
+        }
+        return java.util.Optional.empty();
+    }
+
+    private static AttributeConverter<Row> createRowDocumentConverter(
+            TableSchema<Row> tableSchema) {
+        return new AttributeConverter<>() {
```
gguptp, line 131: i think it should be 

```
return new AttributeConverter<Row>()
``` 
instead

ddebowczyk92: Hey @gguptp! It's diamond operator - Java infers the generic type  from the return type declaration, `<Row>` would be redundant here

gguptp: Yep, im aware, but for some reason, it was throwing an error in my ide

gguptp (approved):

ddebowczyk92:
@ferenc-csaky can we have this merged?

flink-connector-aws/flink-connector-dynamodb/src/test/resources/create-table.sql
```diff
@@ -15,6 +15,7 @@ CREATE TABLE dynamo_db_table (
     `some_time` TIME,
     `some_timestamp` TIMESTAMP(3),
     `some_timestamp_ltz` TIMESTAMP_LTZ(5),
+    `some_row` ROW<myField INT, myOtherField String>,
```
davidradl, line 18: I see you are testing different topologies - I suggest checking a ROW of ROWs as well.

ddebowczyk92: @davidradl thanks for looking in, I addressed your suggestion

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverter.java
```diff
@@ -84,18 +91,63 @@ private StaticTableSchema.Builder<RowData> addAttribute(
...
+                                                    .toExternalOrNull(
+                                                            fieldGetter.getFieldOrNull(rowData)))
+                            .setter(((rowData, t) -> {}));
+                    buildRowAttributeConverter(field.getDataType())
+                            .ifPresent(a::attributeConverter);
+                });
+    }
+
+    private java.util.Optional<AttributeConverter> buildRowAttributeConverter(DataType dataType) {
+        if (LogicalTypeRoot.ROW.equals(dataType.getLogicalType().getTypeRoot())) {
```
ferenc-csaky: Enums can be compared with the equality operator, so I'd rather use that instead of `.equals`. Same in L117 and L159

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverter.java
```diff
@@ -84,18 +91,63 @@ private StaticTableSchema.Builder<RowData> addAttribute(
...
+                                                            field.getDataType())
+                                                    .toExternalOrNull(
+                                                            fieldGetter.getFieldOrNull(rowData)))
+                            .setter(((rowData, t) -> {}));
+                    buildRowAttributeConverter(field.getDataType())
+                            .ifPresent(a::attributeConverter);
+                });
+    }
+
+    private java.util.Optional<AttributeConverter> buildRowAttributeConverter(DataType dataType) {
```
ferenc-csaky: Let's add a qualified import for `Optional`.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverter.java
```diff
@@ -104,8 +156,36 @@ private <T> EnhancedType<T> getEnhancedType(DataType dataType) {
...
     }
+
+    private TableSchema<Row> buildRowTableSchema(DataType dataType) {
+        StaticTableSchema.Builder<Row> builder = TableSchema.builder(Row.class);
+        AttributeConverterProvider newAttributeConverterProvider =
+                new ArrayAttributeConverterProvider();
+        builder.attributeConverterProviders(
+                newAttributeConverterProvider, AttributeConverterProvider.defaultProvider());
+
+        DataTypes.Field[] fields = DataType.getFields(dataType).toArray(new DataTypes.Field[0]);
```
ferenc-csaky: Any reason for the array conversion? `getFields` returns an `ArrayList`, so calling `.get` on the list element has no downside, hence I do not see a reason to make it an array.

ferenc-csaky (changes requested):

ferenc-csaky (approved):
LGTM, thanks for iterating on it! Will merge if CI is green.

flink-connector-aws/flink-connector-dynamodb/src/test/resources/create-table.sql
```diff
@@ -15,6 +15,8 @@ CREATE TABLE dynamo_db_table (
...
+    `some_row_in_row` ROW<myField INT, ROW<innerRowField INT, innerRowOtherField String>>,
```
ferenc-csaky: The test fails, cause this col definition is invalid. I think what we miss is the field name for the inner row, so it should be:
```SQL
`some_row_in_row` ROW<myField INT, myInnerRow ROW<innerRowField INT, innerRowOtherField STRING>>
```

ferenc-csaky (changes requested):

ferenc-csaky (approved):

PR #247, opened 2026-05-12
[FLINK-39660][Connectors/Kinesis] Fix Netty event loop threads blocking behavior and concurrent activation race in FanOutKinesisShardSubscription

chenylee-aws:
## Purpose of the change

Fix FLINK-39660 (https://issues.apache.org/jira/browse/FLINK-39660): a cascading failure in the Kinesis EFO subscription lifecycle. Under transient disruptions (downstream slow consumer, GC pause, brief network error) the connector enters a self-sustaining retry storm that stalls consumption across many shards, leaks HTTP/2 streams, and saturates the JVM common ForkJoinPool. See the JIRA for observed symptoms and an illustrative stack trace.

  This PR addresses the identified issues in FanOutKinesisShardSubscription:
  
  - Netty event loop no longer blocks on record processing. Switched from blocking eventQueue.put() to non-blocking offer(), with a prefetch-based pull model.
  - Concurrent activations on the same shard are prevented. Replaced the subscriptionActive-based guard (which only blocked after onSubscribe) with an identity-based guard on shardSubscriber. Activations during the pending window are now correctly rejected.
  - Identity-based disposal. All four error paths (responseHandler.onError, .exceptionally, timeout callback, Subscriber.onError) funnel through disposeIfActive(subscriber); only the winning path cleans up. Losing paths no-op instead of corrupting state.
  - Added close() invoked from FanOutKinesisShardSplitReader.close() for clean shutdown; pending subscriptions are disposed and further activations are rejected.
  - KinesisStreamsSource now passes user-configured properties to the Netty HTTP client instead of an empty AttributeMap, so http-client.max-concurrency and related settings actually apply.
  

## Verifying this change

  - Tests use a scripted AsyncStreamProxy fake that lets each test deterministically drive onSubscribe, onNext, onComplete, handler.onError, and future.completeExceptionally signals. Avoids Thread.sleep in favour of Awaitility polling.
  - Load-tested against a high throughput stream for days before and after the change. With the fix: no concurrent activations observed, no zombie subscriptions, no ForkJoinPool common-pool saturation in thread dumps, no Netty event-loop blocking on put. Incoming and outgoing throughput of the stream stay closely to each other. 
 
## Note
This PR should also address the following issues 
[FLINK-39660](https://issues.apache.org/jira/browse/FLINK-39660)
[FLINK-39540](https://issues.apache.org/jira/browse/FLINK-39540)
[FLINK-38814](https://issues.apache.org/jira/browse/FLINK-38814)
[FLINK-37949](https://issues.apache.org/jira/browse/FLINK-37949)
[FLINK-37648](https://issues.apache.org/jira/browse/FLINK-37648)
[FLINK-34071](https://issues.apache.org/jira/browse/FLINK-34071)

### DIFF
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
@@ -32,6 +32,7 @@ import software.amazon.awssdk.services.kinesis.model.SubscribeToShardEvent;
 import java.time.Duration;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 
 import static org.apache.flink.connector.kinesis.source.config.KinesisSourceConfigOptions.EFO_CONSUMER_SUBSCRIPTION_TIMEOUT;
 
@@ -44,6 +45,7 @@ public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
     private final AsyncStreamProxy asyncStreamProxy;
     private final String consumerArn;
     private final Duration subscriptionTimeout;
+    private final ScheduledThreadPoolExecutor timeoutScheduler;
 
     private final Map<String, FanOutKinesisShardSubscription> splitSubscriptions = new HashMap<>();
 
@@ -56,6 +58,15 @@ public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
         this.asyncStreamProxy = asyncStreamProxy;
         this.consumerArn = consumerArn;
         this.subscriptionTimeout = configuration.get(EFO_CONSUMER_SUBSCRIPTION_TIMEOUT);
+        this.timeoutScheduler =
+                new ScheduledThreadPoolExecutor(
+                        1,
+                        r -> {
+                            Thread t = new Thread(r, "subscription-timeout-scheduler");
+                            t.setDaemon(true);
+                            return t;
+                        });
+        this.timeoutScheduler.setRemoveOnCancelPolicy(true);
     }
 
     @Override
@@ -71,6 +82,7 @@ public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
         boolean shardCompleted = event.continuationSequenceNumber() == null;
         if (shardCompleted) {
             splitSubscriptions.remove(splitState.getShardId());
+            subscription.close();
         }
         return new RecordBatch(event.records(), event.millisBehindLatest(), shardCompleted);
     }
@@ -85,7 +97,8 @@ public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
                             consumerArn,
                             split.getShardId(),
                             split.getStartingPosition(),
-                            subscriptionTimeout);
+                            subscriptionTimeout,
+                            timeoutScheduler);
             subscription.activateSubscription();
             splitSubscriptions.put(split.splitId(), subscription);
         }
@@ -94,6 +107,7 @@ public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
     @Override
     public void close() throws Exception {
         splitSubscriptions.values().forEach(FanOutKinesisShardSubscription::close);
+        timeoutScheduler.shutdownNow();
         asyncStreamProxy.close();
     }
 }
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
@@ -93,6 +93,7 @@ public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
 
     @Override
     public void close() throws Exception {
+        splitSubscriptions.values().forEach(FanOutKinesisShardSubscription::close);
         asyncStreamProxy.close();
     }
 }
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
@@ -47,7 +47,6 @@ import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
@@ -70,15 +69,7 @@ public class FanOutKinesisShardSubscription {
                     TimeoutException.class,
                     IOException.class,
                     LimitExceededException.class);
-    private static final ScheduledExecutorService TIMEOUT_SCHEDULER =
-            new ScheduledThreadPoolExecutor(
-                    1,
-                    r -> {
-                        Thread t = new Thread(r, "subscription-timeout-scheduler");
-                        t.setDaemon(true);
-                        return t;
-                    });
-
+    private final ScheduledExecutorService timeoutScheduler;
     private final AsyncStreamProxy kinesis;
     private final String consumerArn;
     private final String shardId;
@@ -100,20 +91,21 @@ public class FanOutKinesisShardSubscription {
     private ScheduledFuture<?> timeoutFuture;
     private FanOutShardSubscriber shardSubscriber;
     private boolean closed = false;
-
-    private volatile StartingPosition startingPosition;
+    private StartingPosition startingPosition;
 
     public FanOutKinesisShardSubscription(
             AsyncStreamProxy kinesis,
             String consumerArn,
             String shardId,
             StartingPosition startingPosition,
-            Duration subscriptionTimeout) {
+            Duration subscriptionTimeout,
+            ScheduledExecutorService timeoutScheduler) {
         this.kinesis = kinesis;
         this.consumerArn = consumerArn;
         this.shardId = shardId;
         this.startingPosition = startingPosition;
         this.subscriptionTimeout = subscriptionTimeout;
+        this.timeoutScheduler = timeoutScheduler;
     }
 
     /** Method to allow eager activation of the subscription. */
@@ -168,7 +160,7 @@ public class FanOutKinesisShardSubscription {
 
             cancelTimeoutFuture();
             timeoutFuture =
-                    TIMEOUT_SCHEDULER.schedule(
+                    timeoutScheduler.schedule(
                             () -> {
                                 String errorMessage =
                                         "Timeout when subscribing to shard "
@@ -423,6 +415,12 @@ public class FanOutKinesisShardSubscription {
         @Override
         public void onComplete() {
             synchronized (lockObject) {
+                if (shardSubscriber != this) {
+                    LOG.warn(
+                            "Ignoring late onComplete for shard {} from a disposed subscriber.",
+                            shardId);
+                    return;
+                }
                 LOG.info("Subscription complete - {} ({})", shardId, consumerArn);
                 shardSubscriber = null;
             }
--- a/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
+++ b/flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
@@ -85,13 +85,14 @@ public class FanOutKinesisShardSubscription {
     private final Duration subscriptionTimeout;
 
     /**
-     * Number of events to keep in flight per subscriber. Pipelining the fetch overlaps the
-     * server's next-event work with the consumer's drain work. Must match the capacity of
-     * {@link #eventQueue}.
+     * Number of events to keep in flight per subscriber. Pipelining the fetch overlaps the server's
+     * next-event work with the consumer's drain work. Must match the capacity of {@link
+     * #eventQueue}.
      */
     private static final int PREFETCH = 2;
 
-    private final BlockingQueue<SubscribeToShardEvent> eventQueue = new LinkedBlockingQueue<>(PREFETCH);
+    private final BlockingQueue<SubscribeToShardEvent> eventQueue =
+            new LinkedBlockingQueue<>(PREFETCH);
     private final AtomicReference<Throwable> subscriptionException = new AtomicReference<>();
 
     // All fields below are guarded by lockObject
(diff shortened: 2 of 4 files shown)


flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -70,22 +70,37 @@ public class FanOutKinesisShardSubscription {
...
+     */
+    private static final int PREFETCH = 2;
+
+    private final BlockingQueue<SubscribeToShardEvent> eventQueue = new LinkedBlockingQueue<>(PREFETCH);
     private final AtomicReference<Throwable> subscriptionException = new AtomicReference<>();
 
-    // Store the current starting position for this subscription. Will be updated each time new
-    // batch of records is consumed
-    private StartingPosition startingPosition;
+    // All fields below are guarded by lockObject
```
pelaezryan, line 89: nit: I could see this comment diverging easily with future changes. New variables are likely to be added below this line. An alternative way to relate the variables to the lock is to prefix/suffix the variable names to include some indicator. (i.e. ``guardedTimeoutFuture``)

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -102,100 +117,132 @@ public FanOutKinesisShardSubscription(
...
+    private void cancelTimeoutFuture() {
+        if (timeoutFuture != null) {
+            timeoutFuture.cancel(false);
+            timeoutFuture = null;
+        }
+    }
+
+    // Must be called while holding lockObject
+    private boolean disposeIfActive(FanOutShardSubscriber subscriber) {
+        if (shardSubscriber != subscriber) {
```
pelaezryan, line 221: do we need to pass in a subscriber and do this validation? Shouldn't there be just the one?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -274,61 +300,52 @@ public SubscribeToShardEvent nextEvent() {
...
-                    startingPosition,
-                    consumerArn);
-            this.subscription = subscription;
-            this.subscriptionState.set(SubscriptionState.SUBSCRIBED);
-            subscriptionLatch.countDown();
+            synchronized (lockObject) {
+                if (shardSubscriber != this) {
+                    // Timeout/error disposed this subscriber and a new one was created before SDK
+                    // called onSubscribe
+                    subscription.cancel();
```
pelaezryan, line 318: Are we sure that the subscription passed in is not in use by the new subscriber?

pelaezryan:
We have the following in fetchrecords. Should we be checking the subscription is closed as well for removing the subscription? There's no way to recover from a ``closed`` subscription without recreating the ``FanOutKinesisShardSubscription``

```
    @Override
    protected RecordBatch fetchRecords(KinesisShardSplitState splitState) {
        FanOutKinesisShardSubscription subscription =
                splitSubscriptions.get(splitState.getShardId());

        SubscribeToShardEvent event = subscription.nextEvent();
        if (event == null) {
            return null;
        }

        boolean shardCompleted = event.continuationSequenceNumber() == null;
        if (shardCompleted) {
            splitSubscriptions.remove(splitState.getShardId());
        }
        return new RecordBatch(event.records(), event.millisBehindLatest(), shardCompleted);
    }
```

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
chenylee-aws: Yes, it's essential. Each error callback (`responseHandler.onError`, `.exceptionally`, timeout scheduler, `Subscriber.onError`) captures its own `FanOutShardSubscriber` in the closure at activation time. By the time the callback fires, `shardSubscriber` may already point at a different subscriber — e.g., an EFO 5-minute `onComplete → activateSubscription` rotation may have  replaced it. Without the identity check, a stale callback would dispose the new subscriber. The check also deduplicates the 3-4 error paths that fire for the same failure (SDK delivers the same error through multiple layers): only the winning path does teardown, the rest no-op cleanly.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
chenylee-aws: Yes. The `Subscription` parameter is the reactive-streams handle for this specific `subscribeToShard` call. It's bound one-to-one with this `FanOutShardSubscriber` instance (the SDK creates a fresh one per call and invokes `onSubscribe` with it exactly once, per Reactive Streams Rule 2.12). The replacement subscriber has its own `subscribeToShard` call producing its own distinct `Subscription`. Cancelling this stale one releases this particular HTTP/2 stream slot without affecting the new subscriber's stream.

pelaezryan (approved):

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -102,100 +118,130 @@ public FanOutKinesisShardSubscription(
...
-                "Activating subscription to shard {} with starting position {} for consumer {}.",
-                shardId,
-                startingPosition,
-                consumerArn);
-        if (shardSubscriber != null
-                && shardSubscriber.getSubscriptionState() == SubscriptionState.SUBSCRIBED) {
-            LOG.warn("Skipping activation of subscription since it is already active.");
-            return;
-        }
+        synchronized (lockObject) {
```
gguptp, line 113: i am wondering if we can make this lock free.

`activateSubscription` is called by `FanoutKinesisShardSplitReader` which is single threaded and holds all the subscriptions. I'm also not sure that if the schedule function of the `TIMEOUT_SCHEDULER` will work as we are expecting or not.

chenylee-aws: The `handleSplitChange` method always starts a new `FanoutKinesisShardSubscription` so there would be no active subscription running and this lock is pretty much a no-op. Besides that, there is no long blocking operation performed under any of the lock blocks so really not expecting any contention.

gguptp (approved):

addu390:
@chenylee-aws Thanks for putting this together, EFO is not usable without these changes today and we have a local build with similar changes to get past it.

Noticed that it requires a review from a maintainer.
@rmetzger No rush, but could you review/approve when you get a chance?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -337,56 +353,89 @@ public void onNext(SubscribeToShardEventStream subscribeToShardEventStream) {
...
             }
+            setSubscriptionException(throwable);
         }
 
         @Override
         public void onComplete() {
-            LOG.info("Subscription complete - {} ({})", shardId, consumerArn);
-            this.subscriptionState.set(SubscriptionState.COMPLETED);
+            synchronized (lockObject) {
+                LOG.info("Subscription complete - {} ({})", shardId, consumerArn);
```
ferenc-csaky, line 424: I think `onComplete` also need to guard against a stale subscriber that `onNext` and `onError` already has in place, to make sure a late `onComplete` for an old subscriber as a result of some unhappy path cannot corrupt subscription state. Something like the below should happen first inside `synchronized`:
```java
if (shardSubscriber != this) {
    LOG.warn("Ignoring late onComplete for shard {} from a disposed subscriber.", shardId);
    return;
}
```

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -337,56 +353,89 @@ public void onNext(SubscribeToShardEventStream subscribeToShardEventStream) {
...
 
-    /** States that the {@code FanOutShardSubscriber} may be in. */
-    private enum SubscriptionState {
-        NOT_STARTED,
-        SUBSCRIBED,
-        COMPLETED
+    public void close() {
+        synchronized (lockObject) {
+            closed = true;
+            if (shardSubscriber != null) {
```
ferenc-csaky, line 434: I'd put `cancelTimeoutFuture();` right after `closed = true` to make sure it will happen.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
```diff
@@ -70,22 +70,38 @@ public class FanOutKinesisShardSubscription {
...
                     IOException.class,
                     LimitExceededException.class);
+    private static final ScheduledExecutorService TIMEOUT_SCHEDULER =
+            new ScheduledThreadPoolExecutor(
+                    1,
+                    r -> {
+                        Thread t = new Thread(r, "subscription-timeout-scheduler");
+                        t.setDaemon(true);
+                        return t;
+                    });
```
ferenc-csaky: I do not believe this `ExecutorService` should be placed as a `static` field here. Making it a `daemon` will not block JVM exit, true, but we cannot monitor it's lifecycle this way and cannot shut it down properly.

Instead I think we can add this as a simple `private final` field into `FanOutKinesisShardSplitReader`, initialize it in the `FanOutKinesisShardSplitReader` constructor, and then we can call a proper shutdown on it in `FanOutKinesisShardSplitReader#close`. Then we can simply pass it to `FanOutKinesisShardSubscription` when the reader initializes it, so it will still be centralized.

After we initialize the `timeoutScheduler` in the reader ctor, I'd also add `timeoutScheduler.setRemoveOnCancelPolicy(true)`.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
ferenc-csaky, line 84: Was not touched by this change, but I think we should add `subscription.close()` after this remove, cause in the `close()` method if this class, it will only close the subscriptions that map still contains.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/test/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscriptionTest.java
ferenc-csaky, line 1: To spare some time and tokens, I already put together a test case to validate, feel free to use it. This code already assumes that `ScheduledThreadPoolExecutor` is managed by `FanOutKinesisShardSplitReader` and only gets passed to `FanOutKinesisShardSubscription` as a ctor param, as I suggested in my other comment.
```java
    @Test
    void onCompleteFromStaleSubscriberDoesNotDisposeActiveSubscriber() throws Exception {
        ScriptedProxy proxy = new ScriptedProxy();
        FanOutKinesisShardSubscription subscription = newSubscription(proxy, LONG_TIMEOUT);

        subscription.activateSubscription();
        ScriptedSubscription s1 = proxy.awaitSubscription();
        s1.onSubscribeDelivered();

        s1.fireHandlerError(new IOException("dispose s1"));
        assertThat(subscription.nextEvent()).isNull();
        ScriptedSubscription s2 = proxy.awaitSubscription();
        s2.onSubscribeDelivered();

        s1.onComplete();

        waitShort();
        assertThat(proxy.subscribeCallCount()).isEqualTo(2);

        s2.deliverEvent(
                SubscribeToShardEvent.builder()
                        .records(record("live"))
                        .continuationSequenceNumber("live-cont")
                        .build());
        SubscribeToShardEvent got = pollEvent(subscription);
        assertThat(got.continuationSequenceNumber()).isEqualTo("live-cont");
    }

    private FanOutKinesisShardSubscription newSubscription(AsyncStreamProxy proxy) {
        return newSubscription(proxy, DEFAULT_TIMEOUT);
    }

    private FanOutKinesisShardSubscription newSubscription(
            AsyncStreamProxy proxy, Duration subscriptionTimeout) {
        ScheduledThreadPoolExecutor timeoutScheduler = new ScheduledThreadPoolExecutor(1);
        timeoutScheduler.setRemoveOnCancelPolicy(true);
        timeoutSchedulers.add(timeoutScheduler);
        return new FanOutKinesisShardSubscription(
                proxy,
                CONSUMER_ARN,
                SHARD_ID,
                StartingPosition.fromStart(),
                subscriptionTimeout,
                timeoutScheduler);
    }
```

ferenc-csaky (changes requested):
Thanks for the fixes, added a couple comments.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
chenylee-aws: Thanks for the suggestion. Moving it to the caller.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
chenylee-aws: Good catch. Applied this change

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscription.java
chenylee-aws: I don't think there's a reachable state where `shardSubscriber == null` but `timeoutFuture != null` since every path besides `onComplete` that nulls shardSubscriber goes through `disposeIfActive`, which already calls `cancelTimeoutFuture()`. For `onComplete`, the timeout should already have been cancelled by `onSubscribe`.  So I'd prefer to keep it as-is and avoid calling `cancelTimeoutFuture()` in too many places

flink-connector-aws/flink-connector-aws-kinesis-streams/src/test/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSubscriptionTest.java
chenylee-aws: Thanks, added it to the tests

ferenc-csaky (approved):
LGTM

PR #252, opened 2026-07-19
[FLINK-40184][Connectors/DynamoDB] Remove DescribeStream inconsistency-resolution check from shard discovery

riyarawat-amazon:

Removes the redundant DescribeStream inconsistency-resolution check from the DynamoDB Streams source enumerator. The check added extra `DescribeStream` calls and could discard an entire discovery cycle, stalling progress.

The enumerator previously detected "closed leaf, no child" inconsistencies in `DescribeStream` output and retried with `exclusiveStartShardId` up to 5 times; if unresolved it returned an empty shard list and early-returned. This is redundant: discovery runs periodically (every 60s) so a transiently-missing child shard is picked up on a subsequent cycle, and parent-before-child ordering is already enforced at assignment time by the parent-finished/absent gate in `SplitTracker`.

JIRA: https://issues.apache.org/jira/browse/FLINK-40184

## Brief change log

- `DynamoDbStreamsSourceEnumerator`: drop the `inconsistencyDetected` early-return in `processDiscoveredSplits`, delete `trackSplitsAndResolveInconsistencies`, and simplify `discoverSplits` to `return streamProxy.listShards(streamArn, null)`
- Delete `SplitGraphInconsistencyTracker` (+ test)
- Remove `ListShardsResult.inconsistencyDetected` (+ test)
- Remove the `DESCRIBE_STREAM_INCONSISTENCY_RESOLUTION_RETRY_COUNT` config option
- Remove `ShardUtils.isShardOlderThanInconsistencyDetectionRetentionPeriod` (+ its 25h retention constant, + test)

## Verifying this change

This change is covered by existing enumerator and split-tracker tests; only the tests covering the deleted code were removed. Existing tests pass.

- Dependencies (does it add or upgrade a dependency): **no**
- The public API, i.e., is any changed class annotated with `@Public(Evolving)`: **no** (removes an internal config option only)
- The serializers: **no**
- The runtime per-record code paths (performance sensitive): **no**
- Anything that affects deployment or recovery: **no**
- The S3 file system connector: **no**

## Documentation

- Does this pull request introduce a new feature? **no**
- If yes, how is the feature documented? not applicable


flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/config/DynamodbStreamsSourceConfigConstants.java
```diff
@@ -45,13 +45,6 @@ public enum InitialPosition {
                     .defaultValue(Duration.ofSeconds(60))
                     .withDescription("The interval between each attempt to discover new shards.");
 
-    public static final ConfigOption<Integer> DESCRIBE_STREAM_INCONSISTENCY_RESOLUTION_RETRY_COUNT =
```
ferenc-csaky, line 48: Removing this is backward-compatible, right? So if someone explicitly overridden this config in their `config.yaml`, after this gets release, this simply gets ignored instead of an invalid config, correct?

Based on the change, it seems like that to me, but I wanna double-check to be sure, cause otherwise we must deprecate instead of remove to be able to release this in a minor version.

ferenc-csaky (approved):
LGTM, added 1 question

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/source/config/DynamodbStreamsSourceConfigConstants.java
gguptp: Yes, this is backward compatible, if this gets released and customer has put this in their config.yaml, it will get ignored

gguptp (approved):

PR #253, opened 2026-07-21
[FLINK-35500][Connectors/DynamoDB] DynamoDB Table API sink fails to delete elements due to key not found

riyarawat-amazon:
## Purpose of the change

When the DynamoDB Table API sink consumes a changelog (CDC) stream containing
DELETE records, it fails with:

```
DynamoDbException: The provided key element does not match the schema
```

because the sink sends the whole item as the delete key instead of only the
primary key. This PR builds the DELETE key from the declared `PRIMARY KEY`
(partition key and optional sort key), leaving PUT (INSERT/UPDATE_AFTER)
unchanged.

## Verifying this change

This change adds unit tests and was verified end to end:

- `RowDataToAttributeValueConverterTest` — `testDeleteOnlyPrimaryKey`,
  `testDeleteOnlyPrimaryKeys`, `testPKIgnoredForInsert`,
  `testPKIgnoredForUpdateAfter`.
- `RowDataElementConverterTest` — DELETE includes only the primary key, DELETE
  without a primary key throws a clear error, and INSERT/UPDATE_AFTER still
  write the full item.
- `DynamoDbDynamicSinkFactoryTest` — primary key is read from the schema,
  composite key order is preserved, `PARTITIONED BY` defaults to the primary key
  when unset, a primary key with more than two columns is rejected, and
  mismatched `PARTITIONED BY` / `PRIMARY KEY` is rejected.
- Manually verified against a real DynamoDB table: a CDC stream of 4 inserts + 1
  delete results in the deleted row being removed. The same job fails with the
  original error on unpatched `main`.

## Significant changes

- *(Bug fix)* DELETE requests now carry only the primary key.
- `PRIMARY KEY` is now required to process CDC DELETE records; a clear
  `TableException` is thrown when it is missing. This is not a breaking change,
  as this scenario previously failed outright.
- When `PARTITIONED BY` is not specified it defaults to the `PRIMARY KEY`, so CDC
  deletes work without an explicit `PARTITIONED BY` clause. When both are
  specified they must reference the same columns.
- Documentation updated for the `PRIMARY KEY` / CDC delete behavior.

Based on the original work in #152 by @robg-eb. The broader `PARTITIONED BY`
deprecation discussed there is intentionally left out of scope.


gguptp (approved):

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSink.java
```diff
@@ -142,6 +149,7 @@ public static class DynamoDbDynamicTableSinkBuilder
         private Properties dynamoDbClientProperties;
         private DataType physicalDataType;
         private Set<String> overwriteByPartitionKeys;
+        private List<String> primaryKey = Collections.emptyList();
```
ferenc-csaky: Do not see the relevance of the default value. The factory will set this properly, which is why it's omitted for all the other fields. I suggest to comply with that.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSink.java
```diff
@@ -77,6 +81,7 @@ protected DynamoDbDynamicSink(
         this.dynamoDbClientProperties = dynamoDbClientProperties;
         this.physicalDataType = physicalDataType;
         this.overwriteByPartitionKeys = overwriteByPartitionKeys;
+        this.primaryKey = primaryKey == null ? Collections.emptyList() : primaryKey;
```
ferenc-csaky: Do not see the relevance of the null check here. This ctor is only called from the builder, which is only called from the factory, which prevalidates these fields, which I guess is why it's omitted for all the other fields. I suggest to comply with that.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
```diff
@@ -48,14 +52,53 @@ public DynamicTableSink createDynamicTableSink(Context context) {
...
+                                    + "table they must reference the same columns, but PARTITIONED "
+                                    + "BY was %s and PRIMARY KEY was %s. Either align them or "
+                                    + "specify only the PRIMARY KEY.",
+                            declaredPartitionKeys, primaryKey));
+        }
+
+        Set<String> overwriteByPartitionKeys = new HashSet<>(declaredPartitionKeys);
+        if (overwriteByPartitionKeys.isEmpty()) {
+            overwriteByPartitionKeys = new HashSet<>(primaryKey);
+        }
```
ferenc-csaky: I think it makes sense to extract these into 2 separate private methods that deals with the validation and only returns the value, so the general property flow is more clean.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
```diff
@@ -48,14 +52,53 @@ public DynamicTableSink createDynamicTableSink(Context context) {
         DynamoDbConfiguration dynamoDbConfiguration =
                 new DynamoDbConfiguration(catalogTable.getOptions(), factoryHelper.getOptions());
 
+        List<String> primaryKey =
+                catalogTable
+                        .getResolvedSchema()
+                        .getPrimaryKey()
+                        .map(UniqueConstraint::getColumns)
+                        .orElse(Collections.emptyList());
```
ferenc-csaky: nit: Collections.emptyList() -> List.of()

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataElementConverter.java
```diff
@@ -37,39 +40,55 @@ public class RowDataElementConverter implements ElementConverter<RowData, Dynamo
...
+    }
+
+    public RowDataElementConverter(DataType physicalDataType, List<String> primaryKey) {
+        this(physicalDataType, primaryKey, false);
+    }
+
+    public RowDataElementConverter(
+            DataType physicalDataType, List<String> primaryKey, boolean ignoreNulls) {
         this.physicalDataType = physicalDataType;
+        this.primaryKey = primaryKey == null ? Collections.emptyList() : primaryKey;
```
ferenc-csaky: Since this already gets the safe primaryKeys, technically the null-check can be omitted, so I suggest to do so, as no extra validation is done for the other fields too.

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverter.java
```diff
@@ -51,15 +53,32 @@ public class RowDataToAttributeValueConverter {
...
+    }
+
+    public RowDataToAttributeValueConverter(DataType physicalDataType, List<String> primaryKey) {
+        this(physicalDataType, primaryKey, false);
+    }
+
+    public RowDataToAttributeValueConverter(
+            DataType physicalDataType, List<String> primaryKey, boolean ignoreNulls) {
         this.physicalDataType = physicalDataType;
+        this.primaryKey = primaryKey == null ? Collections.emptyList() : primaryKey;
```
ferenc-csaky: Same as my other comments about the null-check

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataToAttributeValueConverter.java
```diff
@@ -68,6 +87,26 @@ public Map<String, AttributeValue> convertRowData(RowData row) {
...
+     *
+     * @param row the row to extract the primary key from
+     * @return a map of the primary key attribute names to their {@link AttributeValue}s
+     */
+    public Map<String, AttributeValue> convertRowDataToKey(RowData row) {
+        Map<String, AttributeValue> item = tableSchema.itemToMap(row, ignoreNulls);
+        Map<String, AttributeValue> key = new LinkedHashMap<>();
+        for (String keyAttributeName : primaryKey) {
+            AttributeValue value = item.get(keyAttributeName);
+            if (value != null) {
```
ferenc-csaky: If for some reason `value` is null here, shouldn't we throw an exception here instead of swallowing it and possibly return an empty or invalid map here?

I guess it will fail on the DynamoDB side anyways, but we can probably pinpoint the anomaly a bit more accurate if we throw an exception here, probably `IllegalArgumentException` or something like that.

ferenc-csaky (changes requested):
Thanks for the change, added some comments.

In general, let's move from JDK8 collection utils from JDK11 ones, so instead of `Collections.singletonList`, `Arrays.asList`, use `List.of`. We finally dropped JDK8 boundaries with Flink 2.x. :)

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSink.java
riyarawat-amazon: Ack

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSink.java
riyarawat-amazon: Ack

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
riyarawat-amazon: Ack, will update

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/DynamoDbDynamicSinkFactory.java
riyarawat-amazon: Ack

flink-connector-aws/flink-connector-dynamodb/src/main/java/org/apache/flink/connector/dynamodb/table/RowDataElementConverter.java
riyarawat-amazon: Ack

riyarawat-amazon:
Thanks for the review! Addressed all comments.

ferenc-csaky (approved):
LGTM!

PR #254, opened 2026-08-17
[FLINK-39137][Connectors/Kinesis] Add configurable interval between non-empty getRecords calls

bube:
## Purpose of the change

Kinesis allows 5 `GetRecords` calls per second per shard. A source that fetches at every opportunity can eat most of that quota, causing other consumers of the same shard to fail with `ProvisionedThroughputExceededException`.

This adds `source.reader.nonempty-records-fetch-interval` to throttle the reader after a fetch that *returned* records — the counterpart to the existing `source.reader.empty-records-fetch-interval`. It is applied per shard.

Defaults to `0 ms`, preserving today's fetch-at-first-opportunity behaviour. Changing the default here would silently throttle existing users.

## Brief change log

- Add `READER_NON_EMPTY_RECORDS_FETCH_INTERVAL`, default `0 ms`.
- `KinesisShardSplitReaderBase` schedules the next fetch after a non-empty batch when the interval is set.
- Document both reader interval options in the SQL connector table. The existing `empty-records-fetch-interval` was undocumented and its description said "in milliseconds" for a `Duration`.
- Sync chinese docs with the english since they are very outdated. A stale translation ticket already exists: https://issues.apache.org/jira/browse/FLINK-25736

## Verifying this change

Added to `KinesisShardSplitReaderBaseTest`:

- `testGetRecordsIntervalForNonEmptySource`: a configured interval is respected.
- `testNoGetRecordsIntervalForNonEmptySourceByDefault`: unset preserves current behavior.

Also verified by hand against localstack on a MiniCluster.

## Significant changes
    - Configuration key descriptions added/updated
    - Added to Table docs, including missing `empty-records-fetch-interval`
    - Synced chinese docs with the english. The zh docs were very outdated. Tested that both english and chinese docs render correctly with ./build_docs.sh from the flink repo
  
## Use of AI assistance

I wrote the implementation. AI assistance (Claude Code 2.1.233) was used to review it and apply the review outcome. The `Duration.ZERO` default in place of a nullable `Long`, the backwards-compatibility test, and the docs. Both commits carry a `Generated-by:` trailer per [ASF policy](https://www.apache.org/legal/generative-tooling.html). I have reviewed
everything and verified it locally.

### DIFF
--- a/docs/content.zh/docs/connectors/table/kinesis.md
+++ b/docs/content.zh/docs/connectors/table/kinesis.md
@@ -409,7 +409,7 @@ Connector Options
       <td>no</td>
       <td style="word-wrap: break-word;">250 ms</td>
       <td>Duration</td>
-      <td>The interval between fetches that returned no records. The interval is applied independently to each shard.</td>
+      <td>The interval to wait on a shard after a fetch returned no records before fetching from it again.</td>
     </tr>
     <tr>
       <td><h5>source.reader.nonempty-records-fetch-interval</h5></td>
@@ -417,7 +417,7 @@ Connector Options
       <td>no</td>
       <td style="word-wrap: break-word;">0 ms</td>
       <td>Duration</td>
-      <td>The interval between fetches that returned records. By default there is no interval, and the next fetch is performed at the first opportunity. The interval is applied independently to each shard, limiting how often this source calls <code>GetRecords</code> on any single shard.</td>
+      <td>The interval to wait on a shard after a fetch returned records before fetching from it again. Defaults to no interval.</td>
     </tr>
     <tr>
       <td><h5>source.shard.get-records.max-record-count</h5></td>
--- a/docs/content.zh/docs/connectors/table/kinesis.md
+++ b/docs/content.zh/docs/connectors/table/kinesis.md
@@ -39,10 +39,72 @@ Dependencies
 
 Kinesis 连接器目前并不包含在 Flink 的二进制发行版中，请查阅[这里]({{< ref "docs/dev/configuration/overview" >}})了解如何在集群运行中引用 Kinesis 连接器。
 
+### Versioning
+
+There are two available Table API and SQL distributions for the Kinesis connector.
+This has resulted from an ongoing migration from the deprecated `SourceFunction` and `SinkFunction` interfaces to the new `Source` and `Sink` interfaces.
+
+The Table API and SQL interfaces in Flink only allow one TableFactory for each connector identifier.
+Only one TableFactory with identifier `kinesis` can be included in your application's dependencies.
+
+The following table clarifies the underlying interface that is used depending on the distribution selected:
+
+<table class="table table-bordered">
+    <thead>
+    <tr>
+      <th class="text-left" style="width: 40%">Dependency</th>
+      <th class="text-center" style="width: 10%">Connector Version</th>
+      <th class="text-center" style="width: 25%">Source connector identifier (interface)</th>
+      <th class="text-center" style="width: 25%">Sink connector identifier (interface)</th>
+    </tr>
+    </thead>
+    <tbody>
+    <tr>
+      <td><code>flink-sql-connector-aws-kinesis-streams</code></td>
+      <td><code>5.x</code> or later</td>
+      <td><code>kinesis</code>(<code>Source</code>)</td>
+      <td><code>kinesis</code>(<code>Sink</code>)</td>
+    </tr>
+    <tr>
+      <td><code>flink-sql-connector-aws-kinesis-streams</code></td>
+      <td><code>4.x</code> or earlier</td>
+      <td>N/A (no source packaged)</td>
+      <td><code>kinesis</code>(<code>Sink</code>)</td>
+    </tr>
+    <tr>
+      <td><code>flink-sql-connector-kinesis</code></td>
+      <td><code>5.x</code> or later</td>
+      <td><code>kinesis</code>(<code>Source</code>), <code>kinesis-legacy</code>(<code>SourceFunction</code>)</td>
+      <td><code>kinesis</code>(<code>Sink</code>)</td>
+    </tr>
+    <tr>
+      <td><code>flink-sql-connector-kinesis</code></td>
+      <td><code>4.x</code> or earlier</td>
+      <td><code>kinesis</code>(<code>SourceFunction</code>)</td>
+      <td><code>kinesis</code>(<code>Sink</code>)</td>
+    </tr>
+    </tbody>
+</table>
+
+{{< hint warning >}}
+Only include one artifact, either `flink-sql-connector-aws-kinesis-streams` or `flink-sql-connector-kinesis`. Including both will result in clashing TableFactory names.
+{{< /hint >}}
+
+These docs are targeted for versions 5.x onwards. The main configuration section targets `kinesis` identifier.
+For legacy configuration, please see [Configuration (`kinesis-legacy`)](#connector-options-kinesis-legacy)
+
+### Migrating from v4.x to v5.x
+
+There is no state compatibility between Table API and SQL API between 4.x and 5.x.
+This is due to the underlying implementation being changed.
+
+Consider starting the job with v5.x `kinesis` table with `source.init.position` of `AT_TIMESTAMP` slightly before the time when the job with v4.x `kinesis` table was stopped.
+Note that this may result in some re-processed some records.
+
 How to create a Kinesis data stream table
 -----------------------------------------
 
-Follow the instructions from the [Amazon KDS Developer Guide](https://docs.aws.amazon.com/streams/latest/dev/learning-kinesis-module-one-create-stream.html) to set up a Kinesis stream.
+Follow the instructions from the [Amazon KDS Developer Guide](https://docs.aws.amazon.com/streams/latest/dev/introduction.html) to set up a Kinesis stream.
 The following example shows how to create a table backed by a Kinesis data stream:
 
 ```sql
@@ -56,9 +118,9 @@ CREATE TABLE KinesisTable (
 PARTITIONED BY (user_id, item_id)
 WITH (
   'connector' = 'kinesis',
-  'stream' = 'user_behavior',
-  'aws.region' = 'us-east-2',
-  'scan.stream.initpos' = 'LATEST',
+  'stream.arn' = 'arn:aws:kinesis:us-east-1:012345678901:stream/my-stream-name',
+  'aws.region' = 'us-east-1',
+  'source.init.position' = 'LATEST',
   'format' = 'csv'
 );
 ```
@@ -66,7 +128,12 @@ WITH (
 Available Metadata
 ------------------
 
-The following metadata can be exposed as read-only (`VIRTUAL`) columns in a table definition.
+{{< hint warning >}}
+The `kinesis` table Source has a known bug that means `VIRTUAL` columns are not supported.
+Please use `kinesis-legacy` until [the fix](https://issues.apache.org/jira/browse/FLINK-36671) is completed.
+{{< /hint >}}
+
+The following metadata can be exposed as read-only (`VIRTUAL`) columns in a table definition. This is only available in the `kinesis-legacy` connector only.
 
 <table class="table table-bordered">
     <thead>
@@ -110,7 +177,7 @@ CREATE TABLE KinesisTable (
 )
 PARTITIONED BY (user_id, item_id)
 WITH (
-  'connector' = 'kinesis',
+  'connector' = 'kinesis-legacy',
   'stream' = 'user_behavior',
   'aws.region' = 'us-east-2',
   'scan.stream.initpos' = 'LATEST',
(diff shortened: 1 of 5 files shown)


jhung-lyft (approved):
LGTM! Nice fix for active shard throttling. Keeping default 0 for compatibility makes sense.

docs/content/docs/connectors/table/kinesis.md
```diff
@@ -404,6 +404,22 @@ Connector Options
       <td>String</td>
       <td>The <code>ReaderType</code> to use for sources (<code>POLLING|EFO</code>).</td>
     </tr>
+    <tr>
+      <td><h5>source.reader.empty-records-fetch-interval</h5></td>
```
davidradl, line 408: you should update the Chinese version also.

docs/content/docs/connectors/table/kinesis.md
```diff
@@ -404,6 +404,22 @@ Connector Options
...
+      <td>optional</td>
+      <td>no</td>
+      <td style="word-wrap: break-word;">250 ms</td>
+      <td>Duration</td>
+      <td>The interval between fetches that returned no records. The interval is applied independently to each shard.</td>
```
davidradl: "The interval between fetches that returned no records."
seems the different to 
"The interval between fetches that returned records"

I suggest keeping the same text everywhere to reduce the possibility of ambiguity

bube: This `<tr>` is for the existing option, `source.reader.empty-records-fetch-interval`, which was undocumented. The option I added is in the `<tr>` below.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/KinesisShardSplitReaderBase.java
```diff
@@ -179,31 +184,47 @@ private boolean skipUntilScheduledFetchTime(KinesisShardSplitState splitState)
     }
 
     /**
-     * Schedules next fetch time, to be called immediately on the result of a fetchRecords() call.
+     * Schedules the next fetch time. To be called immediately after a fetchRecords() call.
```
davidradl: this reads strangely "Schedules the next fetch time". We are scheduling the next fetch not the fetch time. 

Would something like this be better as the top level summary of the method: 
`Schedules the next fetch at the appropriate time in the future. `

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/KinesisShardSplitReaderBase.java
```diff
@@ -179,31 +184,47 @@ private boolean skipUntilScheduledFetchTime(KinesisShardSplitState splitState)
...
+     * @param recordBatch record batch returned by fetchRecords()
      */
     private void scheduleNextFetchTime(KinesisShardSplitState splitState, RecordBatch recordBatch) {
         if (recordBatch == null || recordBatch.getRecords().isEmpty()) {
+            long scheduledGetRecordTimeMillis = scheduleAt(splitState, emptyRecordsIntervalMillis);
+            if (LOG.isDebugEnabled()) {
+                LOG.debug(
+                        "Fetched zero records from split {}, scheduling next fetch at {}",
+                        splitState.getSplitId(),
+                        new Date(scheduledGetRecordTimeMillis).toInstant());
```
davidradl: does this date format in a readable way?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/KinesisShardSplitReaderBase.java
```diff
@@ -179,31 +184,47 @@ private boolean skipUntilScheduledFetchTime(KinesisShardSplitState splitState)
...
-     * scheduled time, fetcher thread will skip fetching (and have small sleep) for the split.
+     * <p>If recordBatch is null or contains no records, the next fetch is scheduled using
+     * emptyRecordsIntervalMillis. Before the scheduled time, the fetcher thread will skip fetching
+     * (and have a small sleep) for the split.
      *
-     * <p>If recordBatch is not empty, next fetchRecords() time is not scheduled resulting in next
-     * fetch on the split is performed at first opportunity.
+     * <p>If recordBatch is not empty, the next fetch is scheduled using
+     * nonEmptyRecordsIntervalMillis. When that interval is zero (the default), no fetch time is
+     * scheduled and the next fetch on the split is performed at the first opportunity.
```
davidradl: Can you confirm that the next opportuanity path is outsider this code and should not be a schedule with 0 internal?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/KinesisShardSplitReaderBase.java
bube: `Instant.toString()` uses ISO-8601 (e.g. 2011-12-03T10:15:30Z). However, I had blindly copied the log message format from the original code which uses the ancient `java.util.Date` class. Instead, I'll use `Instant.ofEpochMilli()` directly. No need to instantiate a `java.util.Date` object for this.

docs/content/docs/connectors/table/kinesis.md
bube: Thanks. I wasn't aware. Now I am :) The chinese docs looks to be very out of sync and seem to be missing a lot of the options and text.

What's the expected scope for this PR? Should I bring over all the missing text (verbatim, in english)? Or, should I only bring over the two options I added in the english docs? In both cases I assume I should file a jira translation ticket, correct?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/KinesisShardSplitReaderBase.java
bube: Confirmed. The "next opportunity" path is outside these code changes. `fetch()` calls `skipUntilScheduledFetchTime()` to check if should skip. `skipUntilScheduledFetchTime()` checks if a next fetch time has been scheduled by `scheduleNextFetchTime()`:
```
if (scheduledFetchTimes.containsKey(splitState)
                && scheduledFetchTimes.get(splitState) > System.currentTimeMillis())
```
Scheduling a 0 (i.e. NOW + 0) would be redundant (and perform slightly worse) because `scheduledFetchTimes.get(splitState) > System.currentTimeMillis()` always would be false. Not setting it allows the method to return quickly already at `scheduledFetchTimes.containsKey(splitState)` returning `false`.

I pinned the existing behaviour with the test `KinesisShardSplitReaderBaseTest#testNoGetRecordsIntervalForNonEmptySourceByDefault`

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/KinesisShardSplitReaderBase.java
bube: I agree that the sentence reads strangely. This method doesn't always schedule a fetch, though. At the default `nonempty-records-fetch-interval` of 0 it schedules nothing on purpose, because `skipUntilScheduledFetchTime()` treats a missing map entry as "fetch now".

That, plus your other comment on line 195:
> Can you confirm that the next opportuanity path is outsider this code and should not be a schedule with 0 internal?

makes me think the naming of methods and fields is an issue. They do not clearly signal that the method only potentially defers the next `fetch()` call, or that *not* putting an entry in the `scheduledFetchTimes` map is a signal for `fetch()` to "fetch now"

How about these re-names to clarify the design an intention of the code:
`scheduleNextFetchTime()` -> `maybeDeferNextFetch()`
`scheduleAt()` -> `deferNextFetchUntil()`
`scheduledFetchTimes` -> `fetchDeferredUntil`

And obviously updating the javadoc :)

bube: What do you think of the renaming and javadocs in 519e44136c5222cb2dab185053fc256c9c21e200, @davidradl?

docs/content/docs/connectors/table/kinesis.md
bube: The zh page was still documenting the legacy connector, so I've copied the English content over rather than just bolting the two new option rows onto an outdated table. And then I plan to file a translation ticket. Does this sound like to right approach to you, @davidradl?

bube:
Thanks for the review, @davidradl. I've updated the PR now and I would appreciate you taking a second look

bube:
There already exists a rather stale translation ticket for the connector table api docs: https://issues.apache.org/jira/browse/FLINK-25736. Will that suffice, @davidradl?

docs/content.zh/docs/connectors/table/kinesis.md
```diff
@@ -39,10 +39,72 @@ Dependencies
...
+    <tr>
+      <td><code>flink-sql-connector-aws-kinesis-streams</code></td>
+      <td><code>4.x</code> or earlier</td>
+      <td>N/A (no source packaged)</td>
+      <td><code>kinesis</code>(<code>Sink</code>)</td>
+    </tr>
+    <tr>
+      <td><code>flink-sql-connector-kinesis</code></td>
+      <td><code>5.x</code> or later</td>
+      <td><code>kinesis</code>(<code>Source</code>), <code>kinesis-legacy</code>(<code>SourceFunction</code>)</td>
```
davidradl, line 77: I am not an expert at this but it looks like the kinesis-legacy supports a higher code level (5 or greater) than the kinesis (4.x). I would expect something called legacy to be associated with lower versions than the non legacy. 

I see there is an ongoing issue (https://issues.apache.org/jira/browse/FLINK-36671), but this naming seems strange to me.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/KinesisSourceConfigOptions.java
```diff
@@ -90,7 +90,19 @@ public enum ConsumerLifecycle {
                     .durationType()
                     .defaultValue(Duration.ofMillis(250))
                     .withDescription(
-                            "The interval in milliseconds between fetches with empty records");
+                            "The interval between fetches that returned no records. The interval "
```
davidradl: The interval between fetches that returned no records. Implies the interval is between 2 fetches that do not return records. I assume we mean this is the interval we wait after we get no records before retrying the fetch (which may or may not get no records). same concern for the other interval.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/KinesisShardSplitReaderBase.java
```diff
@@ -179,31 +183,48 @@ private boolean skipUntilScheduledFetchTime(KinesisShardSplitState splitState)
...
+     * When that interval is zero (the default), nothing is deferred and the next fetch on the split
+     * is performed at the first opportunity.
      *
-     * @param splitState splitState on which the fetchRecords() was called on
-     * @param recordBatch recordBatch returned by fetchRecords()
+     * @param splitState split state the fetchRecords() call was made for
+     * @param recordBatch record batch returned by fetchRecords()
      */
-    private void scheduleNextFetchTime(KinesisShardSplitState splitState, RecordBatch recordBatch) {
+    private void maybeDeferNextFetch(KinesisShardSplitState splitState, RecordBatch recordBatch) {
```
davidradl: nit: maybe is a strange method name. I wonder if the method should be called getNextFetchDeferInterval.
Have it return the millis and if it is non 0 then  issue deferNextFetchUntil(splitState, deferInterval);

Not a necessary change - but thought I would share.

docs/content.zh/docs/connectors/table/kinesis.md
bube: I copied this over verbatim from the english docs. I think this makes sense. The rendered table looks like this:

| Dependency | Connector Version | Source connector identifier (interface) | Sink connector identifier (interface) |
|:---|:---:|:---:|:---:|
| `flink-sql-connector-aws-kinesis-streams` | `5.x` or later | `kinesis`(`Source`) | `kinesis`(`Sink`) |
| `flink-sql-connector-aws-kinesis-streams` | `4.x` or earlier | N/A (no source packaged) | `kinesis`(`Sink`) |
| `flink-sql-connector-kinesis` | `5.x` or later | `kinesis`(`Source`), `kinesis-legacy`(`SourceFunction`) | `kinesis`(`Sink`) |
| `flink-sql-connector-kinesis` | `4.x` or earlier | `kinesis`(`SourceFunction`) | `kinesis`(`Sink`) |

It states that the old `SourceFunction` and `SinkFunction` implementations of the SQL connector were simply called `kinesis` in version 4.x, and in version 5.x they were renamed to `kinesis-legacy` and a new pair of `Source` and `Sink` based implementations were introduced under the name 'kinesis'.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/KinesisShardSplitReaderBase.java
bube: I like it. Improves the readability. Done

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/config/KinesisSourceConfigOptions.java
bube: I've updated the text now.

bube:
Thanks for the thorough review, @davidradl. I've addressed your comments now.

ferenc-csaky (approved):

PR #213, opened 2025-06-06
[FLINK-37908] Resolve EFO-enabled KinesisStreamsSource initialization failure on automated restarts

digitalpoetry:
## Purpose of the change

Resolve race condition on automated restarts of Flink apps with an EFO-enabled KinesisStreamsSource.

## Verifying this change

This change is a trivial rework / code cleanup without any test coverage.

This change is aligned with a previously implemented fix for the same issue observed within the `FlinkKinesisConsumer` connector: 

- https://issues.apache.org/jira/browse/FLINK-24431
- https://github.com/apache/flink/pull/17417

## Significant changes
*(Please check any boxes [x] if the answer is "yes". You can first publish the PR and check them afterwards, for convenience.)*
- [ ] Dependencies have been added or upgraded
- [ ] Public API has been changed (Public API is any class annotated with `@Public(Evolving)`)
- [ ] Serializers have been changed
- [ ] New feature has been introduced
  - If yes, how is this documented? (not applicable / docs / JavaDocs / not documented)

PR #255, opened 2026-08-19
[FLINK-37908][Connectors/Kinesis] Self-heal EFO consumer on restart races and external deletion

aqibmehmoodDH:
## Purpose of the change

Related to https://issues.apache.org/jira/browse/FLINK-37908, which already has an open fix (#213) for the register/deregister race by making `deregisterStreamConsumer()` a permanent no-op. This PR takes a different approach that also closes a couple of related gaps found while investigating the same area, so I'm submitting it as an alternative for maintainers to weigh against #213 rather than assuming it should replace it.

Three related problems in the `JOB_MANAGED` EFO consumer lifecycle:

1. **Register-time race**: `StreamConsumerRegistrar#registerStreamConsumer()` catches `ResourceInUseException` on conflict but never records `consumerArn` or verifies the consumer actually reached `ACTIVE`. A fast restart (any restart — task failure, redeploy, autoscale, not necessarily Kinesis-related) can land on a consumer that's still `DELETING` from the previous attempt's own deregistration, silently hand off an ARN that isn't actually usable, and leave `consumerArn` null so this attempt's own later deregistration becomes a no-op.
2. **No self-healing on missing consumer**: `KinesisStreamsSource#getConsumerArn()` retries `DescribeStreamConsumer` on `ResourceNotFoundException` but never attempts to recreate the consumer. Since nothing else in the connector re-registers on a task-level restart (the enumerator that owns registration isn't recreated by a pipelined-region restart), a consumer that goes missing for any reason — the race above, or external deletion — stays missing forever from the connector's perspective.
3. **Unbounded, zero-backoff retry, and a shard/consumer mixup**: `FanOutKinesisShardSubscription`'s recoverable-exception handling retried indefinitely with no backoff and no ceiling, so a permanently broken subscription (e.g. its consumer no longer exists) spun retrying forever with no visible task failure. Separately, its `ResourceNotFoundException` handling didn't distinguish "the shard is gone" (expected, e.g. resharding — should mark the split complete) from "the consumer is gone" (should fail the task so it can restart and self-heal) — both surface as the same exception type, and a raw `ResourceNotFoundException` is caught by `KinesisShardSplitReaderBase` and treated as the former in both cases, silently abandoning the shard when it's actually the latter.

## Brief change log

- Added `StreamConsumerRegistrar#ensureActiveConsumer()`: describes the consumer; waits out an in-progress `DELETING` before creating a replacement; registers if missing; tolerates a concurrent `ResourceInUseException` on register as another caller/subtask winning the same race (converges by continuing to poll for `ACTIVE`, like everyone else); returns a **verified** `ACTIVE` ARN or throws after a bounded number of attempts. Used by both `registerStreamConsumer()` (so `consumerArn` is always correctly recorded) and `KinesisStreamsSource#getConsumerArn()` for `JOB_MANAGED` (so a missing consumer gets recreated instead of endlessly re-described). `SELF_MANAGED` behavior is unchanged — it still fails fast and never registers on the user's behalf.
- Added bounded, full-jitter exponential backoff and a max-attempts ceiling to `FanOutKinesisShardSubscription`'s recoverable-exception retry, so a permanently broken subscription eventually throws instead of looping forever.
- `ResourceNotFoundException` found anywhere in the cause chain (not just an exact top-level match — in practice it's frequently delivered wrapped, e.g. in `CompletionException`) is now checked against the exception message to distinguish "no such consumer" (wrapped in `KinesisStreamsSourceException`, fatal, triggers a task failure/restart) from "no such shard" (re-thrown raw as before, so `KinesisShardSplitReaderBase` still marks the split complete as it should for a genuinely closed shard).

## Verifying this change

Added unit tests to `StreamConsumerRegistrarTest` (waits out `DELETING` before registering; tolerates a concurrent `ResourceInUseException` on register; `consumerArn` is recorded even on the conflict path) and `FanOutKinesisShardSubscriptionTest` (backoff is actually delayed, not immediate; eventually throws after the attempt ceiling; counter resets after a successful subscribe; a consumer-not-found `ResourceNotFoundException` is wrapped rather than rethrown raw, while a shard-not-found one still isn't).

Also verified end-to-end against a live Kinesis stream (outside this repo's test suite): registering under real concurrent parallelism (10 parallel subtasks racing to register, one wins, the rest converge cleanly), waiting out a real in-progress deregistration before re-registering, and externally deregistering a running job's consumer and observing it fail fast at the next forced resubscription, restart, self-heal by registering a brand-new consumer, and resume reading — full recovery in ~16 seconds with no manual intervention.

## Significant changes


flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/StreamConsumerRegistrar.java
```diff
@@ -143,6 +144,90 @@ public void deregisterStreamConsumer() {
...
+                                + maxAttempts
+                                + " attempts.",
+                        null);
+            }
+
+            sleepUninterruptibly(computeBackoffMillis(baseDelay, maxDelay, attempt));
+        }
+    }
+
+    /** Full-jitter exponential backoff: random duration between 0 and min(base*2^(n-1), max). */
```
tamizhgeek, line 214: I would use software.amazon.awssdk.core.retry.backoff.FullJitterBackoffStrategy instead of building our own backoff strategy

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/StreamConsumerRegistrar.java
```diff
@@ -143,6 +144,90 @@ public void deregisterStreamConsumer() {
...
+            }
+
+            if (status == ConsumerStatus.ACTIVE) {
+                return consumerArn;
+            }
+
+            if (status == null) {
+                try {
+                    LOG.info("Registering stream consumer - {}::{}", streamArn, consumerName);
+                    kinesisStreamProxy.registerStreamConsumer(streamArn, consumerName);
```
tamizhgeek, line 182: Its a bit weird that we try to register the consumer here, but then fail it at line 198 without even checking if this attempt has succeeded. I would either move the check for attempts >= maxAttempts to the top of the method, or once registered here, check if its successful and return the consumer ARN.

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/StreamConsumerRegistrar.java
```diff
@@ -143,6 +144,90 @@ public void deregisterStreamConsumer() {
...
+    }
+
+    /** Full-jitter exponential backoff: random duration between 0 and min(base*2^(n-1), max). */
+    private static long computeBackoffMillis(Duration baseDelay, Duration maxDelay, int attempt) {
+        long exponential = baseDelay.toMillis() * (1L << Math.min(attempt - 1, 20));
+        long capped = Math.min(exponential, maxDelay.toMillis());
+        return ThreadLocalRandom.current().nextLong(capped + 1);
+    }
+
+    private static void sleepUninterruptibly(long millis) {
```
tamizhgeek, line 221: I think this should be called sleepInterruptibly?

flink-connector-aws/flink-connector-aws-kinesis-streams/src/main/java/org/apache/flink/connector/kinesis/source/reader/fanout/FanOutKinesisShardSplitReader.java
```diff
@@ -42,6 +42,12 @@
  */
 @Internal
 public class FanOutKinesisShardSplitReader extends KinesisShardSplitReaderBase {
+
+    // See FanOutKinesisShardSubscription#maxRecoverableAttempts.
+    private static final int MAX_RECOVERABLE_SUBSCRIBE_ATTEMPTS = 100;
+    private static final Duration RECOVERABLE_SUBSCRIBE_BASE_BACKOFF = Duration.ofMillis(500);
+    private static final Duration RECOVERABLE_SUBSCRIBE_MAX_BACKOFF = Duration.ofSeconds(30);
+
```
tamizhgeek, line 50: If these are constants (not configurable) why do we need to set them in here instead of directly having it on the FanOutKinesisShardSubscription ?

